PR #124, opened 2017-03-02
Add handling for replay of chunked encoding. Closes #121

### LINKED ISSUE #121: Extraneous characters display in replay possibly introduced by wget
I have encountered an issue where extraneous characters are being displayed for simple WARCs in replay. For example, https://github.com/oduwsdl/ipwb/blob/master/ipwb/samples/warcs/froggie.warc.gz shows extraneous 'ac' characters before and '0' after the content. This content is not on the live web capture at http://www.cs.odu.edu/~mkelly/semester/2017_spring/remoteFroggie.html .

The salam WARCs in the core sample data set also have extraneous characters.

Upon further inspection of the froggie warc, this content is inside the WARC, so it's not introduced by ipwb. This may be due to wget's WARC implementation introducing the additional characters, the latter produces with wget 1.18 on macOS.

For froggie I used the command:
`wget -pH --warc-file=froggie http://www.cs.odu.edu/~mkelly/semester/2017_spring/remoteFroggie.html`

machawk1:


ipwb/replay.py
```diff
@@ -255,13 +255,33 @@ def show_uri(path, datetime=None):
...
     return resp
 
 
+def extractResponseFromChunkedData(data):
+    chunkDescriptor = -1
+    retStr = ''
+
+    (chunkDescriptor, rest) = data.split('\n', 1)
+    while chunkDescriptor.strip() != '0':
+        chunkDecFromHex = int(chunkDescriptor, 16)  # Get dec for slice
```
ibnesayeed: The `chunkDescriptor` line is not always just a purely hexadecimal number. If there are some extensions used, it may have some more stuff after a semicolon. If we are only interested in the number of octets for now then at least the `chunkDescriptor` should be split on `;` and the first value should be passed to the `int()` typecasting function or else be ready to face a `ValueError`. By extracting chunk size from chunk descriptor line, we can then set the termination condition of the while loop to compare with integer 0 instead of string 0.

[From Wikipedia:](https://en.wikipedia.org/wiki/Chunked_transfer_encoding#Format)

> Each chunk starts with the number of octets of the data it embeds expressed as a hexadecimal number in ASCII followed by optional parameters (chunk extension) and a terminating CRLF sequence, followed by the chunk data. The chunk is terminated by CRLF.

> If chunk extensions are provided, the chunk size is terminated by a semicolon and followed by the parameters, each also delimited by semicolons. Each parameter is encoded as an extension name followed by an optional equal sign and value. These parameters could be used for a running message digest or digital signature, or to indicate an estimated transfer progress, for instance.

machawk1: Alternatively, we can check if ';' is present in the line. If so, split on it then use the first value as the chunk size. Casting this to an `int` instead of stripping would be better as well -- I'll give you that.

ibnesayeed: I would perhaps do something like this:

```python
chunkSizeHex = chunkDescriptor.split(';')[0].strip()
chunkSizeDec = int(chunkSizeHex, 16)
```

ibnesayeed: Checking the presence of `;` is not needed, the split will still work if the splitter is not present and return only one item in the list as a whole.

machawk1: Ah, ok, I was not sure of that. I recall some languages puking if the character to split is not present. Python is apparently not one of them.

Issue #249, opened 2017-08-14
Issue with getCDXJLinesWithURIR() on Linux


machawk1:
Testing 433940671322b0d685fea3aa560a53fc576bf465 on Ubuntu 16.04

`ipwb index ipwb/samples/indexes/froggie.tar.gz | ipwb replay`

...then access main replay page and click on single capture shown displays "Error" in browser with terminal reporting `getCDXJLinesWithURIR() takes exactly 2 arguments (1 given)`. This is caused by some previous refactoring and the removal of a default value for said function, instead checking if unset within the function and assigning there.

Issue #247, opened 2017-08-09
ipfsapi and ipfs mismatch reports 404 when trying to shut down daemon


machawk1:
Related to approach used to close #241.

Using ipfs v0.4.9 and ipfsapi v0.4.2 (which aligns to ipfs v0.4.9), when the ipfs daemon shutdown button is pressed in the ipwb web UI, the play system reports

`404 Client Error: Not Found for url: http://127.0.0.1:5001/api/v0/shutdown?stream-channels=true`

and does not shut down the daemon.

There are currently catches in place to account for ipfsapi<0.4.2, which does not support the `ipfsapi.shutdown()` command, as the ipfs daemon to which it corresponds (<v0.4.9) does not support the `ipfs shutdown` semantics.

This mismatch between the daemon the and Python API is inevitable, so catch this error and run, throw the exception, and execute the logic to kill the process.

PR #250, opened 2017-08-14
Call getCDXJLinesWithURIR with consistent number of parameters for #249

machawk1:


machawk1:
Note that I forgot to repull master here but was working on the same code basis with only changes described in the ticket. Don't delete this branch yet until the corresponding ticket (#247) is resolved.

Issue #214, opened 2017-06-29
`unsupported locale setting` in Linux on replay


machawk1:
Upon replaying anything any URI-M, the datetime conversion method `datetimeToRFC1123()` is required to set the locale, which it does via `locale.setlocale(locale.LC_TIME, 'en_US')`. This is not a valid locale on Ubuntu 16.04 with the closest variant through `locale -a` being `en_US.utf8`. `en_US` is valid on macOS per `locale -a` but its utf8 variant is `en_US.UTF-8`, along with `en_US`. This subtle difference in the UTF8 variants might cause an issue. Given we are sorting 14-digit datetimes, LC_ALL might be sufficient and more compatible across platforms.

machawk1:
Closed with 4066a3bbfcc60e51f69929f46ba7c712ae100a6d

PR #252, opened 2017-08-14
Try utf8 US locale if enUS is unavailable (Linux). For #214

machawk1:


machawk1:
Closes #214. Manually verified with locale Ubuntu 16.04

Issue #235, opened 2017-07-17
Change default replay listening IP to 0.0.0.0


ibnesayeed:
In order for the service to be accessible from anywhere other than localhost, the service should be listening to a non-loopback interface, preferably on the multicast address `0.0.0.0`. Currently, it is set to `127.0.0.1` as default and not changeable (which is a bug). While making this change, it is also important to DRY the code around it for so that any run-time change is reflected throughout.

machawk1:
Preliminary tests are showing SW not behaving as expected on 0.0.0.0 compared to localhost or 127.0.0.1.

machawk1:
Closing this issue. I am aware that there are redundancies and some remnant hard-coded references, which will be fixed for (#22).

ibnesayeed:
> Preliminary tests are showing SW not behaving as expected on 0.0.0.0 compared to localhost or 127.0.0.1. SW in Chrome 59 reports, "ServiceWorker registration failed: DOMException: Only secure origins are allowed" per https://www.chromium.org/Home/chromium-security/prefer-secure-origins-for-powerful-new-features .

I am aware of this issue, but it can easily be solved. What hostname is being reported in generated URIs is separate from what interface the daemon is listening on. Supply two configuration options, one for listening interface (that defaults to `0.0.0.0`) and one for the hostname (that defaults to `localhost`).

machawk1:
@ibnesayeed I don't follow. The hostname:port is how the user currently accesses their local replay system. What is the "listening interface"?

ibnesayeed:
> @ibnesayeed I don't follow. The hostname:port is how the user currently accesses their local replay system. What is the "listening interface"?

When a service is running on a machine, it can be reached from a network using an IP address and a port number. Basically, the IP address is to locate the host machine while the port number is to locate a process on that host. However, a host machine can have multiple network interfaces that can be listed using `ifconfig` command in Linux machines. A loop back interface is a special network interface that only works on the same machine and does not listen to any traffic from outside. It is accessible from the hostname `localhost` and often mapped to the IP address `127.0.0.1` (with a masking of `255.0.0.0`). It is possible for an application to specify, which of the many network interfaces available on the machine (along with a specific port) should be used to communicate with it. However, if the application wants to listen to all the network interfaces, it can use another special interface available at `0.0.0.0`.

[...]

Issue #148, opened 2017-04-12
If ipfsapi port is changed in ipfs config, ipwb replay reports status as not running


machawk1:
Related to #22 and #147.

Issue #22, opened 2016-05-24
Host and port are hard-coded


machawk1:
The README now supplies a command to specify the ipfs daemon to use a different port. There is no guarantee that it will be at localhost and the default port but this data is hard-coded into `indexer.py`. Parameterize these values to be customizable at runtime for `indexer.py`. Also, potentially define these defaults in an external file.

ibnesayeed:
Do you imagine more parameters that might be passed? If there are more than a couple then it might be a good idea to use some flag library such as [argparse](https://docs.python.org/3/library/argparse.html).

machawk1:
Yes, I do, and I think argparse is the way to go. Our currently logic simply checks for argument count but this ought to be revised to attribute the WARC path to a parameter name (first value w/o a - or -- in argparse setup params). What other arguments do you imagine?

ibnesayeed:
If we are going to use argparse then we don't have to worry about all the possible arguments just yet, we can add more arguments as we need them. Since the order of the arguments doesn't matter, it wont brake anything.

machawk1:
Because ipwb can be installed via source or pip, it's probably best to store a config elsewhere on the system. ~/.ipfs/config is JSON. From there, we can read the IPFS API (and other IPFS) ports but potentially also extend it with ipwb options. Currently, there are also other entries for things related to but not core to ipfs like ipns, so it wouldn't be too far fetched to also store ipwb settings there for the ipwb replay script to r/w.

machawk1:
This is also now causing false positives in the check for the daemon status in #147 . At the very least as a smart default, the replay system ought to consult the ~/.ipfs/config to override the hard-coded 5000 and 5001 values in ipwb.

ibnesayeed:
Should this ticket be closed now?

machawk1:
@ibnesayeed I am testing with some additional host/port supplied by the user to ensure they are used. If it works, I will close this ticket per #400.

PR #326, opened 2017-12-09
Change to use loopback for #235

machawk1:
@ibnesayeed Please check this. PR'd so TravisCI can run tests.


machawk1:
This is what you suggested in #235, @ibnesayeed , where you said, "Basically, you can just hard code app.run(host="0.0.0.0")".

Please suggest an alternative.

machawk1:
I made this change, @ibnesayeed, just waiting on Travis again. The issue of customization is already described in #22 and #148 and is not yet enabled. Please check on the above ticket to let me know if it is satisfactory to close this PR and #235.

ibnesayeed:
> This is what you suggested in #235, @ibnesayeed , where you said, "Basically, you can just hard code app.run(host="0.0.0.0")".
>
> Please suggest an alternative.

Yes, what I said is accurate. Basically, when you create a service (with the intent to listen on on network interfaces), you listen to `0.0.0.0`, but when you connect to a service, you can't use this non-resolvable address, instead used loopback address, localhost, or any other interface where the service is listening to.

ipwb/util.py
```diff
@@ -26,7 +26,7 @@
 from ipfsapi.exceptions import ConnectionError
 
 
-IPFSAPI_IP = '127.0.0.1'
+IPFSAPI_IP = 'localhost'
```
ibnesayeed: Using `localhost` here is equally as valid as `127.0.0.1`. This means, while this change is supposed to work as expected, this is not a **REQUIRED** change for this purpose.

machawk1: localhost makes it consistent with IPWBREPLAY_IP, which is now a bit of a misnomer.

Thanks for the quick feedback.

PR #341, opened 2017-12-19
Removing hardcoded origin from the ServiceWorker

### LINKED ISSUE #337: Localhost is hard-coded in the service worker
https://github.com/oduwsdl/ipwb/blob/cc80a38ddcacf4482215cd93a344daedba60a432/ipwb/serviceWorker.js#L74

ibnesayeed:
This would fix #337.


machawk1:
Tested and works for me. Thanks, @ibnesayeed.

Issue #342, opened 2017-12-20
Remove duplicate endpoints with similar logic


ibnesayeed:
Currently, `replay.py` has two methods `showMemento` and `showMementoAtDatetime` that are being served at `/memento/<datetime>/<urir>` not at `/<datetime>/<urir>` respectively. The latter has duplicate behavior with less functionalities. I think the later should be removed completely, but before that we can set a redirect from one to the other to make sure other things are not breaking. More discussion on this can be found in #339. Current SW implementation actually performs this rerouting on the client side anyway, but for Reconstructive (#338) to be used efficiently, we need to remove this duplicate behavior.

machawk1:
@ibnesayeed Can #339 go forward without this change?

ibnesayeed:
I am yet to see what was the webui.js related issue that was caused due to a little change in the replay.py, so it is difficult to comment on that. If we really want to make it work, perhaps a lot of custom logic can be added in the service worker to make it work, but:

* that is not guaranteed to work
* it will be an even bigger pain to work with and maintain
* fixing the remaining issues in the reconstructive to deal with situations it is failing in right now would be difficult, and
* this endpoint should go away nonetheless

So, I don't really see a good reason to invest in causing more mess rather than taking the opportunity to clean up the code.

ibnesayeed:
I looked at the code and workflow further and realized that in the current implementation `/<datetime>/<urir>` endpoint is only intended to be used by the navigational requests. For everything else (embedded resources that are not under special endpoints) the browser might think they are coming from `/<datetime>/<urir>`, but they are actually being served from `/memento/<datetime>/<urir>` using a plain rerouting logic:

https://github.com/oduwsdl/ipwb/blob/8d7f59e21060cab27cc8dcf79dbcf802ef19259e/ipwb/serviceWorker.js#L72-L74

So, the behavior for now is, if the navigational page (main/root memento page) at that datetime is not available then it will return a 404, but every other resource is almost guaranteed to have silently redirected to the closest memento while completely masking this act from the client/user. This is a very naive and confusing implementation which does not add any real value.

In the new Reconstructive implementation if we add an exclusion for navigational fetches and deal with them differently, then we will either end up duplicating the whole logic for this new pattern or will lose the opportunity to extract datetime from the referrer. Also, such exclusion would mean that we wont be able to leverage reconstructive features for adding banners or performing other rewriting tricks such as to make hyperlinks work as expected.

[...]

Issue #340, opened 2017-12-19
[wish] option to force 404 on any resources not in cdxj


akavel:
Would it be possible to have an option in ipwb (via serviceWorker?) which would force any URLs not found in cdxj to fail/404? So that I could be sure that what I see is built without any external online resources, so I can check if it's readable as such; in other words: I could assess if it's "backed-up well enough for me".

As a workaround, I imagine I could possibly setup some proxy to force no connection to the Internet? or just turn off my WiFi?
 
(Somewhat related to #335)

ibnesayeed:
The intent of the [Reconstructive](https://github.com/oduwsdl/reconstructive) ServiceWorker is to prevent from any live leaks. This means that IPWB (that now uses Reconstructive) will make sure that non-local resources are served from the archive only and not from the live web. Currently, there are some implementation issues, but those are known bugs that will eventually be fixed.

machawk1:
@akavel I believe this is what already occurs. For example, if I

```sh
ipwb index ipwb/samples/warcs/5mementos.warc | ipwb replay
```

then `curl -i http://localhost:5000/20140114100000/memento.us/`, I get an HTTP 200 response, as that's one of the mementos available. However, if I change that command to `curl -i http://localhost:5000/20140114100009/memento.us/` I get the following response:

```html
HTTP/1.0 404 NOT FOUND
Content-Type: text/html; charset=utf-8
Content-Length: 611
Link: <memento.us/>; rel="original", <http://localhost:5000/timemap/link/memento.us/>; rel="timemap"; type="application/link-format", <http://localhost:5000/timemap/cdxj/memento.us/>; rel="timemap"; type="application/cdxj+ors", <http://localhost:5000/20130202100000/memento.us/>; rel="first memento"; datetime="Sat, 02 Feb 2013 10:00:00 GMT",    <http://localhost:5000/20161231110001/memento.us/>; rel="last memento"; datetime="Sat, 31 Dec 2016 11:00:01 GMT"
Server: InterPlanetary Wayback Replay/0.2017.12.12.1843
Date: Tue, 19 Dec 2017 23:06:18 GMT

[...]

PR #339, opened 2017-12-18
Utilizing the reconstructive module in the ServiceWorker

ibnesayeed:
It implements #338 which potentially take care of #335 and #337.

I have added adequate inline documentation for customization. Any generic logic that can go in the reconstructive module should be added there while minimizing customization. There are some known issues in the reconstructive repository which will be fixed soon.

Issue #301, opened 2017-12-01
Allow user to specify non-14-digit datetimes in URI


machawk1:
e.g., GET on `/2016/http://example.com` would return only list of mementos with interpreted date (all from the year of 2016). Currently a non-14-digit date simply returns an interface with all mementos for the URI-R listed as well as an abbreviated TimeMap in the link header.

- [x] Perform sanity check on 14-digit dates (e.g, prevent invalid 20172004101112)
- [x] Resolve partial 14-digit dates
  - [x] Just year (4-digit), e.g., 2014
  - [x] Year and month, e.g., 201409 (September 2014)
  - [x] Year, month, and day, e.g., 20140910 (September 10, 2014)
  - [x] Year, month, day, and hour, e.g., 2014091002 (September 10, 2014 2 am)
  - [x] Year, month, day, hour, and minute, e.g., 201409100215  (September 10, 2014 2:15 am)
  - [X] Year, month, day, hour, minute, and second, e.g., 20140910021539  (September 10, 2014 2:15 am and 39 seconds)
- [ ] Allow datetimes beyond 14-digits (refer to WARC/1.1 spec, which ISO date?) (#283) 
- [x] Prevent non-14-digit dates (e.g., 2014) from resolving to 0-padded invalid dates (20140000000000)

ibnesayeed:
Memento endpoint should return only the closest memento. What you are describing here should be implemented as datetime-based slicing on TimeMap. There are plans to support similar functionality in MemGator too.

machawk1:
Correct, this ticket appeals to the slicing feature and not the `/memento/` endpoint, per above.

ibnesayeed:
I think I am confused with the purposes of `showMementoAtDatetime` and `showMemento`.

In my opinion, sliced TimeMaps should be served at `/timemap/<format>/<[start-datetime]:[end-datetime]>/<urir>`.

machawk1:
`showMementoAtDatetime()` handles requests for `/20171201094013/http://example.com`.

It's really a wrapper to the generic `show_uri()`.

`showMemento()` does the "closest datetime" resolution, i.e., requests for `/memento/*/http://example.com`.

There is definitely some duplicate code and redundant functionality, which stems from progressively adding the functionality.

The sliced TimeMap functionality also needs to serve the web UI when a 404 for a URI-M is encountered. I think the `/timemap/` endpoint ought to be isolated from this.

[...]

PR #343, opened 2017-12-20
Unifying memento endpoint by adding a redirect

ibnesayeed:
This implements #342 for testing the implications of a redirect from one memento endpoint to the other and potentially removing one later.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -437,7 +437,7 @@ def getCompleteURI(uri):
 def showMementoAtDatetime(urir, datetime):
     urir = getCompleteURI(urir)
     datetime = datetime.ljust(14, '0')
-    return show_uri(urir, datetime)
+    return redirect("/memento/{0}/{1}".format(datetime, urir), code=301)
 
 
 @app.errorhandler(Exception)


machawk1:
This change has some undesired side-effects. For example, running `ipwb index ipwb/samples/warcs/5mementos.warc | ipwb replay` then accessing a URI-M with a datetime in the URI-M _close_ to another resolves to the close memento. This is incorrect and should return a 404. Repeating the procedure from the current master branch then tweaking the 14-digit datetime in the URI-M to dereference a memento that does not exist returns a 404, as expected.

screen shot 2017-12-19 at 9 10 18 pm

ibnesayeed:
I don't think it is an undesired effect? All the usual archival replay systems do the same thing. They return a 404 only if there are no captures for the requested URI-R or the closest capture is actually a recorded 404. In every other case, they just redirect to the closest memento.

machawk1:
Despite my reluctance to return a 3XX when a URI-M with a URI-R/datetime pair aren't available instead of a 404 NOT FOUND in the replay, this change will facilitate #301. This change also has ramifications on #340.

With that said, is there any way for us to express in the ultimate representation of the root memento that a forward occurred? HTTP 3XX are often automatically followed  by a browser whereas the redirect would be more evident with a non-browser-based User-Agent like curl.

ibnesayeed:
> With that said, is there any way for us to express in the ultimate representation of the root memento that a forward occurred? HTTP 3XX are often automatically followed by a browser whereas the redirect would be more evident with a non-browser-based User-Agent like curl.

This seemingly simple question is more involved than it feels it is. A composite memento is made up of dozens and some times hundreds of embedded resources. How many redirection actions do we want to report? Redirection to the closest mementos happen all the time, especially for the embedded resources that are often several folds more than root HTML pages. Unless entire page is captured at once and rewritten to point to corresponding datetime values of every embedded resource before storing (along the lines of how archive.is does it), there is no good way of knowing absolute final non-redirecting URI-Ms for every resource from the client side in advance. In a normal archival browsing, we request an HTML page with our desired datetime which then redirects us to the closest root memento which then requests every embedded resource with the same datetime as the served (and potentially redirected) root memento. These requests then silently get redirected to the nearest available captures of each embedded resource. This lazy datetime resolution is almost essential to function archival replay, especially when we strive for no-rewriting for the replay sake.

If we do want to express that a redirect happened, the best way would be to take note of the requested datetime on each navigation request and save it in the localstorage, then add a banner on the root memento to tell what was the requested datetime, how long before or after the served memento was captured and how much is the temporal spread of various embedded resources with respect to the requested datetime or the served root memento datetime.

machawk1:
@ibnesayeed In addition to the banner route, which is a good idea, I would like this information expressed in a header somewhere so UA's that aren't a browser w/ localStorage support are still able to get this information if redirect are followed (e.g., curl's -L flag).

ibnesayeed:
User agents that are not web browsers usually do not follow redirects automatically, so the client has full control of rejecting unwanted redirects or responses that are beyond the tolerance limit temporal distance. In fact, due to the stateless nature of HTTP, the client knows more than what would server know. We can cook up a custom header to express the [positive or negative] time difference, but that will only be reliable on per-requests and response cycle. Any successive redirects will not have the trace of the originally requested datetime. There are situations when there will be more than one redirects to reach to a terminal memento. The client however (browser with localstorage or any other HTTP library or tool) will know the intended original request datetime, and can always calculate the difference from the ultimate terminal response's `Memento-Datetime` header.

ibnesayeed:
Steps to perform test on this PR in Docker as of now.

```shell
$ cd /tmp && git clone https://github.com/oduwsdl/ipwb.git && cd ipwb
$ git checkout -b reconstructive-test # Create a test branch from the master
$ git merge origin/issue-338 # Merge reconstructive integration changes in the test branch
$ git merge origin/issue-23 # Merge changes for the outfile support in indexing for easier testing in Docker
$ mkdir -p /tmp/data/{warc,cdxj,ipfs}
$ wget -P /tmp/data/warc/ https://github.com/ibnesayeed/acrts/raw/master/assets/acrts.warc.gz
$ docker image build -t ipwb .
$ docker container run -it --rm -v /tmp/data:/data ipwb ipwb index -o /data/cdxj/acrts.cdxj /data/warc/acrts.warc.gz
$ docker container run -it --rm -v /tmp/data:/data -p 5000:5000 ipwb ipwb replay /data/cdxj/acrts.cdxj
```

Head over to the http://localhost:5000/ then the follow the link to the archived copy of the [ibnesayeed.github.io/acrts/](http://localhost:5000/memento/20170206002337/ibnesayeed.github.io/acrts/) then scroll the page a little and click on the `By JS On Click` block to get all green on the test page!

ipwb-reconstructive-docker-acrts

ibnesayeed:
Closing this PR as it was only for testing purpose while this change has already been merged as part of #339. However, we will keep the related #342 issue open until that is resolved.

PR #346, opened 2017-12-20
Add scheme to original URI. Closes #295.

### LINKED ISSUE #295: Sanitize URIs in TimeMaps
It is a good idea to accept requests with or without the protocol part in the URI-R, but when generating TimeMaps we should add those in every place where it makes sense.

machawk1:
Ping @ibnesayeed 

= Future Work =
Read the scheme from URI-Ms to see if all URI-Rs in each are https and if so, use https in original as well.


ibnesayeed:
Generally speaking, we should always keep the original URI of each capture in the index under the JSON block, but so far we short-circuited to just the SURT entry as a prefix outside the JSON block. I would store both `originalUrl` and `scheme` to avoid an index with lossy information as well as to make URI-M formation quick.

machawk1:
That might be a better solution -- to store the original URI. Explicitly including `scheme` seems redundant if we are listing `originalURI`.

ibnesayeed:
Yes, that would be redundant, but I thought if parsing scheme is something we are going to do for every URI-M each time when it is to be included in a TM, then we might just do it once at indexing time and avoid repeated parsing. However, on a second thought, it might be too much book keeping, especially when we also think about adding stripped `www` when relevant. I will certainly not like to have fields for all such attributes.

PR #356, opened 2017-12-24
Added outfile support for index, closes #23, validates #355

### LINKED ISSUE #23: Allow the user to specify an outfile other than stdout from indexer for CDXJ data
...and maybe show periodic progress, as the procedure can take awhile.

Related to #22 in the need to add in `argparse`.

ibnesayeed:
This should close #23 and make Docker documentation change #355 valid when merged.

### DIFF
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -28,7 +28,8 @@ def checkArgs_index(args):
         compressionLevel = 6  # Magic 6, TA-DA!
 
     indexer.indexFileAt(args.warcPath, encKey, compressionLevel,
-                        args.compressFirst, outfile=args.outfile, debug=args.debug)
+                        args.compressFirst, outfile=args.outfile,
+                        debug=args.debug)
 
 
 def checkArgs_replay(args):
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -28,7 +28,7 @@ def checkArgs_index(args):
         compressionLevel = 6  # Magic 6, TA-DA!
 
     indexer.indexFileAt(args.warcPath, encKey, compressionLevel,
-                        args.compressFirst, debug=args.debug)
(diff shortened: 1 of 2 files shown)


ibnesayeed:
This seamlessly performs a sorted merge if the outfile has some CDXJ lines already.

ibnesayeed:
Currently, we are opening supplied index file in `a+` mode so that we first read any contents if the file exists (a file with that name will be created if not already). We populate cdxjLines array with the line we read from the existing file so that it is merged with the new stuff that is yet to be indexed. We keep the handle of that open file until it is time to write the merged content to it. We then truncate the index file and write merged index lines to it.

An alternate approach would be to open the file in read mode first (if any) then close it and populate the array. Later, open the file in write mode just before writing.

@machawk1 any prefs between the two approaches?

machawk1:
I'll defer the style of file i/o to your, @ibnesayeed . One approach requires we only have a single file handler but keeps it open through a potentially long process whereas the second has two short-lived handlers but has other issues of other data being written to the target file if written from another process in the meantime.

With that in-mind, I think I prefer the first version but will still defer.

ibnesayeed:
Yeah, there are certainly pros and cons of either approaches, that's why I put this discussion up. Today I did make a change though. Now, we truncate the existing file just before we are ready to write things. Earlier, I implemented it in a way that I was truncating it immediately after reading the content. I made this change to prevent some data loss (for as long as we can). Imagine, someone is running the indexer on new stuff while passing an existing index file to incrementally add to it. However, if the process dies (due to any interrupt or exception) before new content was indexed, we would have lost all the existing indexes in the supplied file, but now with this deferred truncation approach, the data has a better chance to survive.

I think we can mark it implemented (after some tests) for now. However, we can keep the two approaches in mind if we encounter any practical issues in the future.

Issue #353, opened 2017-12-23
We need to periodically tag releases


ibnesayeed:
While we are following a rolling release style model currently, we should tag the repo with a version number after certain big changes.

ibnesayeed:
This will be helpful to release tagged images on DockerHub rather than just the `latest` one based on the bleeding edge master.

ibnesayeed:
Perhaps we can tag the repo every time we bump the version for PyPi.

machawk1:
Often I will fix a bug that warrants a new version but isn't significant enough to be a "big change". Given how rapidly the code is evolving, I think it's redundant to have to add a tag for each tweak. We want the user on the bleeding edge master.

machawk1:
I have been bumping the version number with each bug fix, so that would be a new tag for every single bug fix.

ibnesayeed:
Well, it does to hurt to tag very often, though you can decide to tag the repo only when you bump the second point version number, not necessarily every third point bugfix bump.

machawk1:
The current versioning is date-based, e.g., v. 0.2017.12.23.1741.

ibnesayeed:
Whatever is the release strategy, I think it would be good to have a periodic tagged release when we feel that enough has been changed since the last release. Alternatively, we can tag the repo on every version bump, but remove last insignificant version tags to reduce the number of tags. This will ensure that a historical record of each release is available under the Releases tab automatically and each version also available in DockerHub's repo.

machawk1:
That seems like a lot of bookkeeping. Is there any way to automate this with, say, having a script that calls the command to publish to pypi and create a new GitHub tag?

machawk1:
There's [this script](https://gist.github.com/bclinkinbeard/1331790) that we could adapt. It might also be good to automate updating the version in `__init__.py` as part of this process.

machawk1:
Ping @ibnesayeed, I have created a script to automate the tag creation and pypi deployment process. Let me know your thoughts and whether we could tie-in a hook to docker hub.

ibnesayeed:
I can configure DockerHub to build on each tag creation.

machawk1:
Can this ticket be closed with the introduction of deploy.sh and a few sample runs that populated the repo's tags/releases?

ibnesayeed:
[...]

PR #357, opened 2017-12-25
Renamed deploy.sh to release.sh and used UTC time

ibnesayeed:
This PR improves #353

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -13,7 +13,7 @@ echo $PYVAR\'$VERSION_STRING\'>'ipwb/'$FILE_NAME
 
 # Push to GitHub
 git add 'ipwb/'$FILE_NAME
-git commit -m "DEPLOY: Bump version for pypi to "$VERSION_STRING
+git commit -m "RELEASE: Bump version for pypi to "$VERSION_STRING
 
 # Create a tag in repo
 TAG_NAME='v'$VERSION_STRING
--- a/release.sh
+++ b/release.sh
@@ -5,7 +5,7 @@ git pull
 
 # Update version in project
 PYVAR="__version__ = "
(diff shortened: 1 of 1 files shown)


ibnesayeed:
I have a different opinion here. The script is not deploying anything, it is releasing a new version instead. That said, it does not have the timezone information in it, so it could be problematic in some situations. Suppose, I visit India (which happens to be 9.5 or 10.5 hours ahead of the Eastern Time) and I publish a new version by running that script. A few minutes later you make some changes and push another release using the same script. This last version number is going to have an earlier time (and perhaps a day earlier too). Even if that situation is far fetched, it illustrates the problem of not making it timezone-neutral. These version numbers are for users, which we don't expect them to have the same TZ as ours or know that which TZ the code was released form.

machawk1:
I agree on the renaming of the script.

How about we pin it where it always uses EST instead of the user's timezone? That's as arbitrary as GMT but it's consistent with every version released so far.

ibnesayeed:
While sticking it to any timezone would be much better than using local time at run time. However, UTC (not GMT) is not as arbitrary as EST. That is a time that is default for the web in many protocols, including HTTP. It does not necessarily attach a geolocation with the time. Sticking it to EST will have issues of a later version counted as an earlier one on the night when clocks go back an hour and if DST is not considered then for a fair part of the year release time will be off by an hour. With that said, I think we should go for UTC. The past releases will become inconsistent, but we should not hurt future ones because of the past, especially when it does not have any serious consequences.

machawk1:
@ibnesayeed I'm conceding to your rationale and have merged this PR.

ibnesayeed:
And for the first time we have dated releases on [DockerHub](https://hub.docker.com/r/oduwsdl/ipwb/builds/).

Issue #369, opened 2018-01-08
Test case suggestion, reassociation of a payload when head is absent


machawk1:
Consider the scenario where a user indexes a WARC to ipwb and gets a CDXJ record with `locator` hashes H/P where H is the IPFS hash of the HTTP header and P the hash of the HTTP entity body.

Another user does the same but gets a `locator` value of L/P where L is a different HTTP header (because it will vary due to differences like the Date header) and the same payload.

The first user transfers the CDXJ to a new machine, shuts the original machine down, and waits a sufficient amount of time for the hashed content to be garbage collected from the IPFS gateway. H will no longer be accessible but P will remain accessible because of the second user still having L/P pinned.

When the user comes back online on a new machine, the H/P value will not be replayable, despite P being available in IPFS and H unavailable. L is still available and might be perfectly suitable for the first user to replay P but the first user is unaware of the L/P CDXJ record.

This would be solvable were IPLD able to support the association, so it's moot to discuss it here, but consider the above scenario with the state of the implementation and test out exactly how ipwb would behave and mitigate accordingly.

machawk1:
Add `ipwb/samples/indexes/froggie_badHeaderHash.cdxj` in 72be79bb624a904ba95fdb46f5ed97b38ff03640 as an example where the HTTP header hash does not resolve, preventing the payload from being displayed.

PR #375, opened 2018-01-30
Fabricate header when only payload is available for #369

machawk1:
@ibnesayeed Check this. There are better ways to do it but this at least gets payloads working again if the header if absent from IPFS.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -525,6 +525,10 @@ def show_uri(path, datetime=None):
     class HashNotFoundError(Exception):
         pass
 
+    payloadCatSuccess = False
+    headerCatSuccess = False
+    payload = ''
+    header = ''
     try:
         def handler(signum, frame):
             raise HashNotFoundError()
@@ -532,8 +536,12 @@ def show_uri(path, datetime=None):
         signal.signal(signal.SIGALRM, handler)
         signal.alarm(10)
 
+        # TODO: Check how payload is populated on failure for optimization
+        #  via removing the need for the payloadCatSuccessVariable
         payload = IPFS_API.cat(digests[-1])
+        payloadCatSuccess = True
         header = IPFS_API.cat(digests[-2])
+        headerCatSuccess = True
 
         signal.alarm(0)
 
@@ -548,8 +556,11 @@ def show_uri(path, datetime=None):
         print(traceback.format_exc())
         print(sys.exc_info()[0])
     except HashNotFoundError:
-        print("Hashes not found")
-        return '', 404
+        if not payloadCatSuccess:
+            print("Hashes not found")
+            return '', 404
+        else:  # payload found but not header, fabricate header
+            print("HTTP header not found, fabricating for resp replay")
     except Exception as e:
         print('Unknown exception occurred while fetching from ipfs.')
         print(sys.exc_info()[0])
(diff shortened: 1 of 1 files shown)


ipwb/replay.py
```diff
@@ -603,6 +614,9 @@ def handler(signum, frame):
 
     resp.headers['Memento-Datetime'] = ipwbConfig.datetimeToRFC1123(datetime)
 
+    if not headerCatSuccess:
+        resp.headers['X-Headers-Generated-By'] = 'InterPlanetary Wayback'
```
ibnesayeed: In this case, the response will not have a `content-type` and `content-encoding` headers because those headers are extracted and added from the original headers:

https://github.com/oduwsdl/ipwb/blob/61ec6a5f643d2659483d5621c61e88f408b0d197/ipwb/replay.py#L603-L606

I think we need to surface these two essential headers (the second one is optional) in the CDXJ file (of which the `content-type` is already stored as `mime_type` field). Then when generating the response, we don't need this `not in` conditional, instead, use them directly from the index.

machawk1: Content-type comes over, as it's pulled from the CDXJ but the encoding issue remains. The test replay has gremlins if the encoding is not specified.

> I think we need to surface these two essential headers

From where? This ticket addresses the issue where the headers are not available. Are you suggesting we include this information in the CDXJ?

machawk1: What do we do when this information (content type, content-encoding) is not available in the CDXJ?

ibnesayeed: > Content-type comes over, as it's pulled from the CDXJ but the encoding issue remains. The test replay has gremlins if the encoding is not specified.

Although we do store the `mime_type` in the index, we do not use it at the time of replay and use the one in the original headers instead. We need to also store the `content-encoding` field in the index (if present) and at the time of replay we need to use these fields from the index rather than the original headers.

> From where? This ticket addresses the issue where the headers are not available.

They are not absent, but took longer time to fetch from the IPFS. At the time of indexing this will be available (MIME certainly, Encoding possibly).

ibnesayeed: > What do we do when this information (content type, content-encoding) is not available in the CDXJ?

`content-type` is mostly available in responses, `content-encoding` is only present when necessary. So, in the replay use them if they are present, otherwise let the user-agent decide how to deal with it. It is the best we can do, because that's how original resource was served when it was live.

machawk1: Right, ok, so we need to actually use the MIME type from the index on replay.

My question was, "What do we do when they are absent?" as in, if the CDXJ gets stripped down to not have this information. What is the best way to still allow the payload to be replay without this information being present?

I think your suggestions here are good and have opened up #376 and #377 (whac-a-mole!)

ibnesayeed: > My question was, "What do we do when they are absent?" as in, if the CDXJ gets stripped down to not have this information. What is the best way to still allow the payload to be replay without this information being present?

"You get what you get and you don't get upset" - DSHR

machawk1: Should we try a best-guess at the content encoding and content type in this situation?

ibnesayeed: No, let the user-agents make the best-guess, as those are more mature and can handle some of these things way better than our naive approach would.

machawk1: The user agent currently displays the encoding metadata...
screen shot 2018-01-30 at 6 40 34 pm

ibnesayeed: If we tried fixing that, we might have ended up in even worse detection. I think once we have this info added to the index, such corner cases will not arise very frequently.

ipwb/replay.py
```diff
@@ -548,8 +550,11 @@ def handler(signum, frame):
         print(traceback.format_exc())
         print(sys.exc_info()[0])
     except HashNotFoundError:
-        print("Hashes not found")
-        return '', 404
+        if payload == '':
```
ibnesayeed: @machawk1, payload could legitimately be empty, so assuming that an empty string means it was payload that was missing may lead to some wrong conclusions while actually the header might have had the issue in being fetched from the IPFS.

machawk1: @ibnesayeed See #378

Issue #105, opened 2017-02-09
Implement Memento RFC7089 endpoints in replay


machawk1:
Current ipwb replay scheme does not provide access to multiple copies of the same URI. We should provide `/memento/`, `/timemap/`, and `/timegate/` endpoints. 

- [X] `/memento/`
- [X] `/timemap/`
- [x] `/timegate/`
- [x] `/memento/*/`

machawk1:
Basic TimeMap endpoint written in issue-105 branch. May need to look into a better way of retaining or at least propagating the scheme from the user's request, at least for rel=original.

machawk1:
Memento endpoint resolution of embedded resources is implemented in https://github.com/oduwsdl/ipwb/commit/c3d068e0c5321cc9468b4a6e1af011067f0bcc58 in issue-88 branch, to be merged.

machawk1:
A memento within the system ought to reporting a TimeMap in the Link header of the HTTP response, e.g., curl -I http://127.0.0.1:5000/20160305192247/cs.odu.edu/~salam/ should have an HTTP response header `Link: http://127.0.0.1:5000/timemap/cdxj/cs.odu.edu/~salam/`

I also currently get an error HTTP 500 trying to access the above TM URI w/ InterPlanetary Wayback 0.2017.05.13.2201

machawk1:
Bumped this priority, as LANL is looking to use ipwb in some way but requires the TimeGate endpoint to be fully implemented.

machawk1:
@ibnesayeed I would appreciate any suggestions you have on approaches in accomplishing this. I have a method to complete this in my head but you tend to give good feedback after it's been implemented, which sometimes requires restructuring.

ibnesayeed:
Implementing TimeGate should not take long as some of the binary lookup and redirect logic is already in place for the `/memento` endpoint. We can reuse that function as it is, but before calling it, we will have to extract value of `Accept-Datetime` header, if present (otherwise default to current time), then convert it to the 14-digit format before calling `showMemento()`. We also need to set some additional response headers such as `Via` when `/timegate` endpoint is hit to advertise the fact that we negotiate the content via `Accept-Datetime` header on TimeGate endpoint.

machawk1:
@ibnesayeed Check out dddb63dbda00f1ff690729539175da58e4399223, please.

machawk1:
Current state:
```sh
[...]

PR #391, opened 2018-06-21
Adds TimeGate endpoint

machawk1:


ipwb/util.py
```diff
@@ -146,10 +146,30 @@ def datetimeToRFC1123(digits14):
         newLocale = 'en_US.utf8'
 
     locale.setlocale(locale.LC_TIME, newLocale)
+
+
+def datetimeToRFC1123(digits14):
```
ibnesayeed: Name of this method should reflect that it is the opposite of `rfc1123ToDigits14()`

machawk1: I agree. Stand by.

machawk1: Updated in latest. Investigating a TravisCI tests issue that may have the unit test names hard-coded. Please continue the review.

ibnesayeed:
To deal with various datetime formats, I would define the following functions:

```python
def parseRFC1123(RFC1123_datetime_string):
  return native_datetime_object

def parse14Digit(14Digit_datetime_string):
  return native_datetime_object

def to14Digit(native_datetime_object):
  return "14Digit_datetime_string"

def toRFC1123(native_datetime_object):
  return "RFC1123_datetime_string"
```

These can be reused in many ways as individual helper functions and also as composite functions for example to convert HTTP DateTime to 14-digit format, either two functions can be called inline in sequence or wrapped in another function as following:

```python
def RFC1123To14Digit(RFC1123_datetime_string):
  return to14Digit(parseRFC1123(RFC1123_datetime_string))
```

machawk1:
@ibnesayeed Ok, that could be a good future optimization for clarity and reuse. Let's get it right first then make it fast/pretty.

ibnesayeed:
Add `Vary: Accept-Datetime` response header to TG endpoint. This is a critical requirement.

machawk1:
🎉 Closes #105

PR #412, opened 2018-07-03
Add semantics/accessibility to main replay UI. Closes #406

### LINKED ISSUE #406: What is the purpose of the input text box on the main page of the replay interface?
_I_ know this but it is not immediately apparent in the interface of how the users should use this text box. Entering a value makes it more clear but it ought to be self-evident in the main interface.

machawk1:
screen shot 2018-07-03 at 4 48 39 pm

### DIFF
--- a/ipwb/webui/index.html
+++ b/ipwb/webui/index.html
@@ -21,7 +21,8 @@ let uris = []
   <div>
     <h1><img src="./webui/logo.png" alt="ipwb" /></h1>
     <form onsubmit="handleSubmit(); return false;">
-      <input type="text" placeholder="http://" id="url" />
+      <input type="text" placeholder="http://example.com" id="url" aria-label="Enter a URI" />
+      <input type="submit" value="Search URL in the archive"/>
     </form>
   </div>
   <footer>
--- a/ipwb/webui/webui.css
+++ b/ipwb/webui/webui.css
@@ -36,3 +36,5 @@ details summary:hover {cursor: pointer;}
 details label {padding: 0; color: #8f1d27; width: 200px; text-align: right; display: block; float: left; margin-right: 1.0em; width: 50%;}
 details label#daemonStatusLabel {height: 10px; margin-top: 0; vertical-align: top;}
 details label#swLabel {height: 2.0em;}
+
+form input[type=submit] {width: 850px; clear: both; margin-top: 0.5em;}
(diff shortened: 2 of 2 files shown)


ibnesayeed:
Make the button look like a primary button with width just enough to fit the text along with some padding.

machawk1:
@ibnesayeed Should we retain center alignment?

ibnesayeed:
> Should we retain center alignment?

Yes! We can take design clues from Google.com or other such sites that have the primary function of the landing page as a search form.

ibnesayeed:
I will make the button bigger in height (to match the height somewhat same as the text field) and increase the size of text on it. Also, I would reduce the size of text in the text input field as the placeholder seems too aggressive.

machawk1:
@ibnesayeed Tweak aesthetics in another ticket. Instead of iterating, please be specific of what you want done in one go.

ibnesayeed:
While we are here talking about the accessibility, perhaps we should also consider the form to work without JS by providing a target to the form and giving text field a name, then issuing a redirect from the server side when the URL is received in a query parameter.

ibnesayeed:
> @ibnesayeed Tweak aesthetics in another ticket. Instead of iterating, please be specific of what you want done in one go.

Appropriate size of text, link, or button are also part of accessibility (though, they can be aesthetics related too). For example, very small text can be hard to read on visual media, and thinner links or buttons can be hard to interact with.

machawk1:
> we should also consider the form to work without JS

I agree. The added button makes this more accessible and will facilitate making the form work sans JS.

Regarding font size: the button is the browser default. It's the input text field that is excessively large, so I do not believe we have an issue with those elements in the text size department. We should verify that there are not sizing issues with the other UI elements on this page.

ibnesayeed:
You could have been a little more patient in merging this because I was not only looking at the code changes, but their effects too, that's why I was responding in pieces and iterating over some changes because you were hot-fixing them (which is good). :)

machawk1:
Perhaps the text input should be smaller (normal size) instead of the button bigger?

I am all for tweaking the design but the requirements of the ticket were met and you seemed to want to tweak the aesthetics further (which is good, but out of scope). :)

ibnesayeed:
> Perhaps the text input should be smaller (normal size) instead of the button bigger?

Since, it is the most important (call for action) element on the page, I think it is perfectly fine to keep the text in the input field a little bigger than normal size (perhaps little smaller than what it is right now) and make the button with comparable padding and text size. Bootstrap buttons or Google's homepage are good places to take some clues.

PR #416, opened 2018-07-04
Allow replay text form to function without JS, closes #414

### LINKED ISSUE #414: Allow URI input form on replay to work without JS
Per #412, we added a button to submit the form, which previously relied on hitting enter with a text field in focus. The conventional submission mechanism is currently suppressed via JS and handled via JS.

This should be more accessible. @ibnesayeed provided a mechanism to accomplish this in #412.

machawk1:


ipwb/replay.py
```diff
@@ -153,6 +153,12 @@ def commandDaemon(cmd):
         return Response('bad command!')
 
 
+@app.route('/memento/*/')
+def showMementosForURIRs_sansJS():
+    urir = request.args.get('url')
+    return redirect('/memento/*/' + urir, code=301)
```
ibnesayeed: I think a `302` status code would be more appropriate here, but I am not too sure about my rationale. Perhaps because we do not want to suggest the client to not use this form of URI again for the query.

ipwb/webui/index.html
```diff
@@ -20,8 +20,8 @@
 <div id="wrapper">
   <div>
     <h1><img src="./webui/logo.png" alt="ipwb" /></h1>
-    <form onsubmit="handleSubmit(); return false;">
-      <input type="text" placeholder="http://example.com" id="url" aria-label="Enter a URI" />
+    <form onsubmit="handleSubmit(); return false;" method="get" action="/memento/*/">
+      <input type="text" placeholder="http://example.com" id="url" name="url" aria-label="Enter a URI" />
```
ibnesayeed: In future we might want to revisit this name if we make the search field more generic and accept queries in natural language like `example.com 3 months ago`.

ipwb/replay.py
machawk1: 301 seems to say better than 302, "Use this other URI instead for now on."

ipwb/replay.py
```diff
@@ -153,6 +153,12 @@ def commandDaemon(cmd):
         return Response('bad command!')
 
 
+@app.route('/memento/*/')
+def showMementosForURIRs_sansJS():
+    urir = request.args.get('url')
```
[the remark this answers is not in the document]
machawk1: Good call. The Flask route might also have a fit with query parameters in URI-Rs within URI-Ms. Some more testing is in order.

ipwb/webui/index.html
machawk1: I agree. Fuzzy/free text queries would be a neat feature to have and I think would be novel as a search method for an archival replay system.

ipwb/replay.py
ibnesayeed: Yes, but that has an implicit assumption that the client should have some memory. Also, each redirect will only instruct about one query at a time rather than an entire class of pattern. That said, I am not too confident about my rationale here.

PR #418, opened 2018-07-04
Handle no mementos condition, supplying 404 and search box. Closes #411

### LINKED ISSUE #411: Empty display when 404 using text input on main replay
Using the search box on the replay page for a URI with 0 mementos returns an empty viewport when using a browser.

Also returns a 200 when viewed with curl. This should probably be a 404, right @ibnesayeed?

Replicate:
1. `ipwb index ipwb/sples/warcs/5mementos.warc | ipwb replay`
1. Enter `example.com` in search box and submit form.

43d1637fbfdf1754e5576ab1f392d3fd346c1e94

machawk1:


ipwb/replay.py
```diff
@@ -672,9 +676,23 @@ def handler(signum, frame):
...
 
+def generateNoMementosInterface_noDatetime(urir):
+    msg = '<h1>ERROR 404</h1>'
+    msg += 'No capture(s) found for {0}.'.format(urir)
+
+    msg += '''
+    <form method="get" action="/memento/*/" style="margin-top: 1.0em;">
+      <input type="text" value="{0}" id="url" name="url" aria-label="Enter a URI" />
+      <input type="submit" value="Search URL in the archive"/>
+    </form>'''.format(urir)
```
ibnesayeed: I am wondering, should we keep this form sans-JS or make it behave the same way landing page does. Alternatively, should we server the original landing page with 404 status along with a clear notice text on it to avoid duplicating logic?

machawk1: I am not sure. I found that I would need to replicate a lot of the logic of the search box on the main page but opted for a simpler JS-less solution here.

ibnesayeed: We might want to revisit it. I can think of the template of landing page having a well-known flash message option that receives custom messages based on certain states.

machawk1: Yes, definitely. The .py code I added has some duplicated logic and a suboptimal name. This fix is to get it working then we can improve it.

ibnesayeed: A lot of duplication has to do with its organic growth. Perhaps separating views from data object and logic is necessary at some point.

Issue #400, opened 2018-06-25
Hardcoded hostname in Link header


ibnesayeed:
https://github.com/oduwsdl/ipwb/blob/a5f7c77daab11255ebde3fbaac1a784e498d8646/ipwb/replay.py#L343-L347

This will cause issues when the service is run with a proper domain name.

machawk1:
Similar to #22

PR #419, opened 2018-07-08
Revert hard-coded localhost to hostAndPort recall for #400

machawk1:


ibnesayeed:
I don't see any immediate issue due to this change. Fixing this was important if it were to run anywhere on a host other than localhost. I think it is time now to have a test deployment on our Docker server.

machawk1:
Thanks for the feedback. We really ought to setup a deployment at a FQDN to evaluate it in a more production-like environment.

Issue #399, opened 2018-06-22
Convert the README.rst to .md to prettify the dockerhub display


machawk1:
https://hub.docker.com/r/oduwsdl/ipwb/

...shows the symbols used for linking and formatting in https://github.com/oduwsdl/ipwb/blob/master/README.rst. RST was originally used for display in pypi. Pypi has since had a UI overhaul. Is the markdown support any better? If so, we might try to move to MD for the sake of readability in dockerhub. Alternatively, we might look to finding a way to ensure the README does not contain the meta content and still use the README.rst.

ibnesayeed:
Looks like the proposal to support Markdown in PyPi is finally implemented.

https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi

machawk1:
@ibnesayeed The README has been updated to MD and displays fine in GitHub and PyPi. Dockerhub seems to choke, however. Can you figure out what, if anything, Dockerhub requires differently than GitHub and PyPi in its MD?

machawk1:
Does an interface exist where I can test how MD will be rendered on Dockerhub?

Go ahead and try to fix them if you can without changing the visual display of the GitHub/PyPi document. If that's not possible (i.e., DockerHub MD restriction prevents some GH-MD feature), let me know.

ibnesayeed:
No, I don't think they have a preview option, but I am sure I can fix it. I will work on this and push changes to the `issues-399` branch. If those changes don't work well, then we can revert them.

machawk1:
@ibnesayeed I recall you mentioning display issues with the README on DockerHub within some other repo. Can you link it here? The line breaking on the DockerHub version of the ipwb README contains odd line breaking, mis-aligned consecutive code blocks, etc.

ibnesayeed:
Line breaking in the DH rendering is a long-known issue https://github.com/docker/hub-feedback/issues/124. If we want to fix that now and make it consistent across all three places, we can just remove all the newlines from the MD source where they are not needed.

I have reported the misaligned consecutive code lines issue https://github.com/docker/hub-feedback/issues/988#issuecomment-403898203.

machawk1:
@ibnesayeed I was wondering what you meant by the first statement, checked out the README.md and noticed superfluous line breaks to a line width (maybe 80-chars?). This may have been the result of the RST->MD conversion script, as I am not prone to manually applying this sort of line breaking to free-form text (code is a different story).

[...]

PR #422, opened 2018-07-08
Fix Markdown formatting issues after conversion from RST

ibnesayeed:
Mainly fixed language annotations in code blocks as noted in #399.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,14 @@ having it work on Python 3 as well (see
 
 The latest release of ipwb can be installed using pip:
 
-```bash
+```
 $ pip install ipwb
 ```
 
 The latest development version containing changes not yet released can
 be installed from source:
 
-```bash
+```
 $ git clone https://github.com/oduwsdl/ipwb
 $ cd ipwb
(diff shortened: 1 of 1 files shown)


machawk1:
I have been aware of the latter (PyPi versioning) issue since we started pushing it there and encountered a date with a leading 0. The leading 0 keeps lexical positioning sorting, so I have retained using that style despite PyPi's behavior.

ibnesayeed:
PyPi should take version number as opaque string though.

machawk1:
I completely agree though maybe they do it on the PEP-440 rationale:

> All numeric components MUST be interpreted and ordered according to their numeric value, not as text strings.

machawk1:
On the contrary, same PEP:

> All release segments involved in the comparison MUST be converted to a consistent length by padding shorter segments with zeros as needed.

ibnesayeed:
I came to know the reasoning behind this because they wanted to ensure `00 == 0` in version numbers as per https://github.com/pypa/setuptools/issues/302#issuecomment-202915790.

PR #431, opened 2018-07-10
Remove fixed-width formatting of README

ibnesayeed:
DH rendering breaks lines when there are newlines in the MD source (as discussed in #399).


machawk1:
@ibnesayeed Does Dockerhub pull from source on a release or a change in the repo? I would like to hold off on pushing another release (even if this is required for DockerHub to regenerate) until we have a functional change.

ibnesayeed:
DH generally updates its README and Dockerfile render based on the last successful build. So, it should be fixed there already. No need for a release.

PR #439, opened 2018-07-13
Rm hard-coded loopback ip

machawk1:
Any reason we should have hard-coded `0.0.0.0` here, @ibnesayeed? This PR removes it to rely on the variable.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -968,7 +968,7 @@ def start(cdxjFilePath=INDEX_FILE, proxy=None):
         print('IPWB replay started on http://{0}:{1}'.format(
             IPWBREPLAY_IP, IPWBREPLAY_PORT
         ))
-        app.run(host='0.0.0.0', port=IPWBREPLAY_PORT)
+        app.run(host=IPWBREPLAY_IP, port=IPWBREPLAY_PORT)
     except gaierror:
         print('Detected no active Internet connection.')
         print('Overriding to use default IP and port configuration.')


ipwb/replay.py
```diff
@@ -968,7 +968,7 @@ def start(cdxjFilePath=INDEX_FILE, proxy=None):
         print('IPWB replay started on http://{0}:{1}'.format(
             IPWBREPLAY_IP, IPWBREPLAY_PORT
         ))
-        app.run(host='0.0.0.0', port=IPWBREPLAY_PORT)
+        app.run(host=IPWBREPLAY_IP, port=IPWBREPLAY_PORT)
```
ibnesayeed: This change will start breaking the replay in Docker, unless [its definition in utils](https://github.com/oduwsdl/ipwb/blob/92028cfce3a462e8da13c436fd2e0d15a4fe381a/ipwb/util.py#L32) is changed to:

```python
IPWBREPLAY_IP = '0.0.0.0'
```

The naming choice of this constant is poor. Currently, the name says it's an IP, but the value associated is hardcoded as `localhost` which is a name, not an IP. I think the reason why you did it that way is to not output IP addresses in generated URIs.

machawk1: Ok. Will this being hardcoded as `0.0.0.0` cause any issues when we deploy ipwb instances?

ibnesayeed: No, this can remain hardcoded to `0.0.0.0`. However, the name `IPWBREPLAY_IP` should be changed to something like `IPWB_HOST` or `IPWBREPLAY_HOST` everywhere.

PR #445, opened 2018-07-19
Mv from pycrypto to pycryptodome. Closes #303

### LINKED ISSUE #303: Please consider migrating from pycrypto to pycryptodome
*pycrypto* doesn't currently build on Windows out of the box, requiring [weird workarounds](https://stackoverflow.com/q/41843266/98528), which can be found only after some googling. This is [reported to *pycrypto*](https://github.com/dlitz/pycrypto/issues/221), but then deferred to the following answer:

> Please see https://github.com/dlitz/pycrypto/issues/238.

There, [one can read](https://github.com/dlitz/pycrypto/issues/238):

> **tl;dr:** PyCrypto is dead. Switch to PyCryptodome (GitHub: https://github.com/Legrandin/pycryptodome PyPi: https://pypi.python.org/pypi/pycryptodome). **Do not report any new bugs here!**
> 
> ----
>
> As you may or may not have noticed the last change to PyCrypto was made **three years ago**.
> 
> As a result the developers don't react to issues on the issue tracker anymore. If you find problems and want them to be fixed consider following:
> 
> As a **user** or **developer** who **uses** software which requires PyCrypto:
> 
> -    Tell this to your application's developer to switch to PyCryptodome and tell them PyCrypto is dead.
> 
> As a **developer** who **develops/maintains software depending on PyCrypto**:
> 
>  -   Consider switching to Pycryptodome - it's a well mantained fork and mostly compatible. If your problems persists report your problem to Pycryptodome's developer.
> 
> As a **developer** who wants to **develop new software**:
> 
>   -  _Do not use Pycrypto!_ Use Pycrpytodome or cryptography. PyCrypto is old and not maintained.
> 

Apart from hopefully fixing the issue on Windows, I imagine using a maintained crypto library could be potentially safer?

machawk1:


machawk1:
The tests fail due to pycryptodome not supporting the naive XOR function:

```py
____________________ ERROR collecting tests/test_memento.py ____________________
ImportError while importing test module '/home/travis/build/oduwsdl/ipwb/tests/test_memento.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_memento.py:4: in <module>
    from ipwb import replay
ipwb/replay.py:45: in <module>
    from Crypto.Cipher import XOR
E   ImportError: cannot import name XOR
```
We need to determine a better rudimentary symmetric encryption function to replace this to merge this PR and close the respective ticket.

ibnesayeed:
Use `Crypto.Util.strxor` instead of `Crypto.Cipher.XOR`.

machawk1:
@ibnesayeed Thanks for that information. As soon as Travis finishes this round of tests with the newly included requirement in test-requirements.txt (with the assumption that it still fails), I will perform some manual tests with your suggestion. Changing function names does not sound like a drop-in replacement. :\

ibnesayeed:
Well, it clearly is not a drop-in replacement if they have some publicly exposed functions renamed or reorganized, but at least they have documented it well.

machawk1:
@ibnesayeed This is now ready to merge pending your review.

ipwb/indexer.py
```diff
@@ -37,7 +37,9 @@
 import requests
 import datetime
 
-from Crypto.Cipher import XOR
+from Crypto.Cipher import AES
+from Crypto.Random import get_random_bytes
```
ibnesayeed: Remove this

ipwb/replay.py
```diff
@@ -42,7 +42,11 @@
 from util import INDEX_FILE
 from requests import ReadTimeout
 
-from Crypto.Cipher import XOR
+from base64 import b64decode
+from Crypto.Cipher import AES
+from Crypto.Util.Padding import unpad
```
ibnesayeed: Remove this

ipwb/replay.py
```diff
@@ -629,14 +634,15 @@ def handler(signum, frame):
                              ' containing decryption key: \n> ')
                 keyString = raw_input(askForKey)
 
-        encryptionMethod = None
-        if jObj['encryption_method'] == 'xor':
-            encryptionMethod = XOR
+        paddedEncryptionKey = pad(keyString, AES.block_size)
+        key = base64.b64encode(paddedEncryptionKey)
+
+        paddedEncryptionKey = base64.b64encode(pad(keyString, AES.block_size))
```
ibnesayeed: Remove this

ibnesayeed:
In the reply we never check for the method used for encryption and assume it is AES, though the indexer stored that property in the CDXJ file.

machawk1:
@ibnesayeed Correct, this is a current assumption. We were checking for xor before but then assigning the cipher to a variable and using that variable for the basis of instantiation. In the future I would like to do this again but given we are only supporting AES at the moment _and_ retaining which encryption method was used, I think it is sufficient to maintain this assumption at replay time.

ibnesayeed:
If in future we extend more encryption methods and someone tries to reply a newer index with an older replay, the server might crash due to incompatible key passed to a different encryption method. I think the safe way would be to put that check in place and return an error to the client while logging the issue as unsupported encryption method if the method is anything but `aes`. This will be forward-safe.

machawk1:
How about we chalk this up as a task for a future ticket? The purpose of this issue was to move away from pycrypto, which we have done.

I think it would be worthwhile to explore providing additional methods beyond AES now that we've become familiar with the API.

ibnesayeed:
Well, it is not stopping us from merging it, but putting that check in place and logging otherwise would be the wise thing to do as I explained a potential reason for that in the previous comment. It is very related, so I would have done in this PR itself, but if you feel like creating a separate ticket and a fresh PR for that then go for it.

machawk1:
Additionally, we currently have no way for a user to specify the encryption method at the time of indexing. I am going to table this for future work, as it will hold up resolving the pycrypto security issue.

ibnesayeed:
> Additionally, we currently have no way for a user to specify the encryption method at the time of indexing.

This is not a very strong reason for a check to be absent in the replay. As long as the generated index has a method annotated, it is considered explicit from the replay perspective, though it was added implicitly during the indexing.

machawk1:
@ibnesayeed Please take a look and comment on #448.

PR #452, opened 2018-07-27
Surfaces redirects from URI-Ms. Closes #451

### LINKED ISSUE #451: Archived HTTP redirects are served as HTTP 200s
Sample WARC: `https://github.com/oduwsdl/ipwb/blob/master/ipwb/samples/warcs/redirect.warc`

`ipwb index ipwb/samples/warcs/redirect.warc | ipwb replay` then `curl -I http://localhost:5000/memento/20180727203127/example.com/`produces a memento with an HTTP 200 instead of the HTTP 302 that exists in the WARC. 

Exhibited behavior reported by @james-e-powell via e-mail.

machawk1:
As documented, the regex to pull the part of the URI-M before the URI-R assumes a 14-digit string. This can be done more reliably but for now it fixes the issue of 3XXs not being replayed.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -655,7 +655,7 @@ def show_uri(path, datetime=None):
     if 'status_code' in jObj:
         status = jObj['status_code']
 
-    resp = Response(payload, status=jObj['status_code'])
+    resp = Response(payload, status=status)
 
     for idx, hLine in enumerate(hLines):
         k, v = hLine.split(': ', 1)
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -651,7 +651,11 @@ def show_uri(path, datetime=None):
     hLines = header.split('\n')
     hLines.pop(0)
 
-    resp = Response(payload)
+    status = 200
+    if 'status_code' in jObj:
+        status = jObj['status_code']
+
+    resp = Response(payload, status=jObj['status_code'])
 
     for idx, hLine in enumerate(hLines):
         k, v = hLine.split(': ', 1)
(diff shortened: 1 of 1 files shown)


ipwb/replay.py
```diff
@@ -684,6 +688,12 @@ def handler(signum, frame):
     # respWithLinkHeader = getLinkHeaderAbbreviatedTimeMap(path, datetime)
     # resp.headers['Link'] = respWithLinkHeader.replace('\n', ' ')
 
+    if status[0] == '3':
+        # Bad assumption that the URI-M will contain \d14 but works for now.
+        uriBeforeURIR = request.url[:re.search(r'/\d{14}/', request.url).end()]
+        newURIM = uriBeforeURIR + resp.headers['Location']
```
ibnesayeed: This will probably fail in many cases. I think the simplistic assumption made here is that extracted `uriBeforeURIR` is going to be a directory URI and stored `Location` header will be a relative path. Both of these are unrealistic assumptions. I think you can simply return the `Location` header without any modification and let the SW deal with it. I am not sure how will it behave in case of a full URIs (probably it will be caught by the SW too).

machawk1: The location header prior to this is a URI-R. Without the prepending of the replay system, the redirect forwards the user to the live Web.

machawk1: uriBeforeURIR ends up being `http://localhost:5000/memento/12345678901234/`.

ibnesayeed: I misunderstood the `uriBeforeURIR` earlier. However, `Location` header is not going to be a full URI all the time, it could very well be a relative or absolute path.

PR #455, opened 2018-07-27
Fixes #454 where an expected tuple being returned a Flask 404 Respons…

### LINKED ISSUE #454: 404s from replay cause "Error" to display and a 500 status code
On replay, changing the URI to a URI-M that does not exist produces an HTTP 500 instead of a 404.

For example: `ipwb index ipwb/samples/warcs/5memento.warc | ipwb replay`, then browsing to http://localhost:5000/memento/20140115101500/urinotinwarc.com results in "Error" being displayed in the browser and an HTTP 500 status code to be returned. This may have to do with @ibnesayeed's recent refactoring of exceptions.

Terminal reports:
```
Could not find com,urinotinwarc)/ in CDXJ at /tmp/9N7S8X.cdxj
'Response' object is not iterable
(<type 'exceptions.TypeError'>, TypeError("'Response' object is not iterable",), <traceback object at 0x103c5ed88>)
```

machawk1:
…e obj caused a type error.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -222,9 +222,11 @@ def resolveMemento(urir, datetime):
     cdxjLinesWithURIR = getCDXJLinesWithURIR(urir, indexPath)
 
     closestLine = getCDXJLineClosestTo(datetime, cdxjLinesWithURIR)
+
     if closestLine is None:
         msg = '<h1>ERROR 404</h1>'
         msg += 'No capture found for {0} at {1}.'.format(urir, datetime)
+
         return Response(msg, status=404)
 
     uri = unsurt(closestLine.split(' ')[0])
@@ -237,9 +239,12 @@ def resolveMemento(urir, datetime):
 
 @app.route('/memento/<regex("[0-9]{1,14}"):datetime>/<path:urir>')
 def showMemento(urir, datetime):
-    (newDatetime, linkHeader, uri) = resolveMemento(urir, datetime)
+    resolvedMemento = resolveMemento(urir, datetime)
 
-    resp = Response()
+    # resolved to a 404, flask Response object returned instead of tuple
+    if type(resolvedMemento) is type(Response()):
+        return resolvedMemento
+    (newDatetime, linkHeader, uri) = resolvedMemento
 
     if newDatetime != datetime:
         resp = redirect('/memento/{0}/{1}'.format(newDatetime, urir), code=302)
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -242,7 +242,7 @@ def showMemento(urir, datetime):
     resolvedMemento = resolveMemento(urir, datetime)
 
     # resolved to a 404, flask Response object returned instead of tuple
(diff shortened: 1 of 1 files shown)


ipwb/replay.py
```diff
@@ -237,9 +239,12 @@ def resolveMemento(urir, datetime):
 
 @app.route('/memento/<regex("[0-9]{1,14}"):datetime>/<path:urir>')
 def showMemento(urir, datetime):
-    (newDatetime, linkHeader, uri) = resolveMemento(urir, datetime)
+    resolvedMemento = resolveMemento(urir, datetime)
```
ibnesayeed: We can perhaps raise a custom `NoMementoFound` error from the `resolveMemento` function and catch it here. Having multiple different return signatures in a function is not a good practice.

machawk1: That sounds like a good idea to improve this. However, lets get this merged in to fix the big issue.

ibnesayeed: Then add a TODO note in the `resolveMemento` function and fix the following line too before merging it:

https://github.com/oduwsdl/ipwb/blob/6b8ac9dd6b38128d7fa0db96f0b4a51bbde2a09a/ipwb/replay.py#L313

machawk1: @ibnesayeed If we use an Exception for this (what you meant by `NoMementoFound` "error"), I think there will be an issue answering the Flask route with the expected response object...unless we have the except return a Flask.response with status of 404 and other attributes about the request. No handling the exception in the...exception handler...sounds like it is not the best way to go.

ibnesayeed: `NoMementoFound` (or something like that) will be a custom exception. The exception will be handled by calling `resolveMemento` in a try block. The `404` response can be generated in the corresponding except block. If there is some information that is only available in `resolveMemento` then it can be placed inside of the error message of the exception.

machawk1: `resolveMemento()` ideally returns `(newDatetime, linkHeader, uri)` tuple. If we throw an exception there as well as handle it within `resolveMemento()`, what do you anticipate `resolveMemento()` returning? Currently it creates a `flask.Response` and returns this. We type check in the caller (`showMemento()` and  `queryTimeGate()`) currently, which I know is a sub-optimal redundancy, but I don't think that the refactoring you suggest will resolve the issue of `resolveMemento()` returning two different types (a `flask.Response` or a `tuple`).

ibnesayeed: The purpose of raising an exception to come out of the `resolveMemento` without any return value as the expected tuple cannot be populated there. We do not want to handle that exception in the same function from where it is being raised. We are already doing that inline handling, which is not good. The exception will be handled in the caller functions (i.e., `showMemento` and `queryTimeGate`).

ibnesayeed: If we do not want to handle those exceptions, we might be able to raise and existing `NotFound` exception from the `werkzeug.exceptions`. I am not sure if Flask will handle it as it bubbles up. If it does on the application level then it should work without explicitly handling it every time `resolveMemento` is called.

Issue #450, opened 2018-07-27
Link on replay homepage omit /memento/ in path, always resulting in an initial redirect


machawk1:
Links to mementos on replay page point to scheme://host/date/URIR instead of scheme://host/memento/date/URIR.

@ibnesayeed Can you think of a reason why we need to link to the URI-M without the `/memento/` part of the path?

ibnesayeed:
Since I have discovered the two endpoints I have always been against this redundancy (#342) which now simply redirects global one to the namespaced one (#343). Once we remove these generated global links from the landing page, we can actually remove that endpoint from the replay.

PR #459, opened 2018-07-28
Remove duplicate memento route

### LINKED ISSUE #342: Remove duplicate endpoints with similar logic

ibnesayeed:
This should fix #342. A related ticket #450 is closed already.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -518,13 +518,6 @@ def getCompleteURI(uri):
     return uri
 
 
-@app.route('/<regex("[0-9]{1,14}"):datetime>/<path:urir>')
-def showMementoAtDatetime(urir, datetime):
-    urir = getCompleteURI(urir)
-    datetime = datetime.ljust(14, '0')
-    return redirect("/memento/{0}/{1}".format(datetime, urir), code=301)
-
-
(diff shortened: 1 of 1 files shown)


machawk1:
@ibnesayeed Travis is reporting that tests are failing with your PR. Can you have a look at this?

ibnesayeed:
I was looking into it. I think I know what's wrong here. A fix is on the way.

machawk1:
@ibnesayeed Once you fix the pycodestyle issue reported, we should be able to merge this.

ibnesayeed:
This looks like passing now. However, in future we need to be vigilant about places where old style memento links might be generated.

Issue #449, opened 2018-07-21
Improve banner design and usefulness


machawk1:
<img width="603" alt="screen shot 2018-07-20 at 9 55 17 pm" src="https://user-images.githubusercontent.com/2514780/43031050-f02863aa-8c67-11e8-8c07-62944961450d.png">

The current replay banner (pictured) has a few issue that should be improved:

- [x] Datetime is an awkward, long, numerical string
- [x] URI-R and datetime are text _editable_ text fields
- [x] Alignment of "Archived at:" and datetime has a lot of awkwardly unused space

I know https://github.com/oduwsdl/reconstructive has much more capability than this. Can we look into taking advantage of this capability, @ibnesayeed?

ibnesayeed:
The improved banner has not been pushed into Reconstructive repo yet. It needs more work to be functional.

machawk1:
Noted @ibnesayeed. Per #144, we are planning on providing a means of navigating between mementos. Perhaps this will be an opportunity to improve the aesthetics and interaction of ipwb's usage of reconstructive.

ibnesayeed:
PR #462 has more than what was enlisted in the OP.

ibnesayeed:
Closing, as this is done!

Issue #144, opened 2017-03-30
Show a banner for inter-capture navigation and interaction w/ ipwb when viewing a memento


machawk1:
Related to #26

Most archival replay systems have a banner splayed across the top of the capture to allow inter-capture navigation. ipwb might benefit from this by also providing additional functionality in this banner to allow interaction with the system from the web interface without going back to the root page. Mock-ups / bulleted list of features to come.

ibnesayeed:
The banner is necessary not only to provide interactivity, but also to display metadata and other relevant information or visualization.

machawk1:
Right, so the banner should contain:

* Navigation to other captures of the same URI-R
* Explicit non-14-digit datetime for capture
* Input box to enter a new URI-R
* Link (maybe a logo) back to the root web UI
* Hide header button/image

As ipwb-specific features in the header:

* ipfs hash of the html page (with option for extended information/hashes of embedded resources)
* Ability to add WARCs/CDXJ to the running ipwb (#119), though this can get messy.

Other suggestions?

ibnesayeed:
> Navigation to other captures of the same URI-R

Some sort of aggregated summary of the captures (preferably some interactive visualization).

> Explicit non-14-digit datetime for capture

I would much prefer a natural language representation of the time using moment.js (which adds tooltip with the actual datetime value).

> Input box to enter a new URI-R (and datetime #111)

Pre-populated with the current URI-R/URI-M.

> Hide header button/image

Yes, and possibly a way to form the URI-M that disables banner automatically.

machawk1:
> natural language representation
That's what I meant: something human-readable. moment.js would be a good way to accomplish this. Other good points as well, @ibnesayeed.

machawk1:
Per IIPC, Perma.cc uses an frames for the banner, which helps to keep content at bay relative to security. This is partially due to the content living on another domain than the other files. Consider this and https://github.com/harvard-lil/warcgames for personal archive contents' security.

ibnesayeed:
Since IPWB uses client-side rerouting using ServiceWorker, which can be coded to never let any external request leak out, so the cross-site attacks would be less likely.

machawk1:
@ibnesayeed I'd like to be a little more sure about this. IIRC, ServiceWorker has issues performing the rerouting within iframes. We don't have any tests for this. We need tests for this.

ibnesayeed:
[...]

PR #462, opened 2018-07-30
Update reconstructive banner implementation

ibnesayeed:
It implements #449 and #144, but the `Link` parsing is not done yet so navigational buttons will not work. However, the banner is ready to function if the values are supplied.

### DIFF
--- a/ipwb/reconstructive-banner.js
+++ b/ipwb/reconstructive-banner.js
@@ -2,7 +2,7 @@
  * [ReconstructiveBanner](https://oduwsdl.github.io/Reconstructive/reconstructive-banner.js) implements `<reconstructive-banner>` [Custom Element](https://html.spec.whatwg.org/multipage/custom-elements.html#custom-elements).
  * It is an unobtrusive archival replay banner to make [mementos](http://mementoweb.org/about/) interactive and surface on-demand metadata about the archived resource.
  * Use it in an HTML page as illustrated below:
- * 
+ *
  * ```html
  * <script src="reconstructive-banner.js"></script>
  * <reconstructive-banner logo-src=""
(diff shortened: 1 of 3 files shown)


ibnesayeed:
Link parsing is implemented as well.

machawk1:
@ibnesayeed The datetime is no longer displayed. Is this intentional?

462c1ddf4e7a288cde40470148c3a9f6d95a69eb

screen shot 2018-07-30 at 11 21 24 am

@ibnesayeed The datetime is not displayed

machawk1:
@ibnesayeed Disregard my above comment. My browser was using a stale SW/reconstruct.

machawk1:
@ibnesayeed Thoughts on changing the logo in the banner to ipwb?

machawk1:
Great work on this, btw!

ipwb/reconstructive-banner.js
```diff
@@ -54,19 +54,25 @@ class ReconstructiveBanner extends HTMLElement {
       if (diff < 0) {
         return 'Capture from the future!';
       }
+      let unit, quotient;
       if (diff >= 31622400000) {
-        return `Captured ${Math.round(diff / 31622400000)} year(s) ago`;
+        unit = 'year';
```
machawk1: We should give semantics to these magic numbers in the future.

ibnesayeed: These numbers are milliseconds in respected bigger units of time that are not going to change. For readability purpose we can define them as constants if you prefer that way.

ibnesayeed:
> Thoughts on changing the logo in the banner to ipwb?

I made it configurable, because it was obvious that we will eventually use the branding of the archive not the banner element. So, go ahead and try it by changing the config in the `serviceWorker.js` file. Logo dimensions might be an issue, but that can be fixed later.

ipwb/reconstructive-banner.js
machawk1: @ibnesayeed Yes, defining them as constants would be an improvement.

ibnesayeed: I have made this change in the upstream repo. Will synchronize here later.

Issue #465, opened 2018-07-30
Proposal for some UI/UX improvements in the replay banner


machawk1:
![banner](https://user-images.githubusercontent.com/2514780/43408616-ccce3a80-93ef-11e8-98f9-e74e042364a6.gif)

In #462 @ibnesayeed made some **_great_** updates to the ipwb replay banner to vastly improve its aesthetics and functionality, but I think there are a few things we can do to make it more useable. I am hoping we can use this ticket for discussing particular improvements. Below I have provided an initial set.

- [x] Highlight forward/back buttons on hover to provide an additional selected context -- redundant of the pointer icon but visually useful.
- [x] Dim forward/backward buttons when not clickable (currently just indicated by no-pointer on hover)
- [X] Prevent some series of clicks from causing the UI elements to appear "selected" (see GIF) (_see resolution below_)
- [ ] Provide some indication with the human-readable date that a hover will reveal the more specific datetime
- [x] Change the Reconstruct icon to ipwb logo
- [x] Add some event (go to ipwb replay home?) when the banner logo is clicked

ibnesayeed:
Most of these are addressed in the upstream repo now.

machawk1:
@ibnesayeed I am looking forward to you merging the upstream Reconstructive changes to ipwb when you get the chance.

ibnesayeed:
> Prevent sone series of clicks from causing the UI elements to appear "selected" (see GIF)

I thought this was handled, is it still happening? If so, then how to reproduce it?

> Provide some indication with the human-readable date that a hover will reveal the more specific datetime

I think that will make the UI noisy. In presence of the natural language time version, highlighting it is not essential.

machawk1:
>  If so, then how to reproduce it?

I can reproduce it with the 5mementos WARC by going to the earliest memento, clicking the right arrow icon 6 times, and noting that the logo highlights.

machawk1:
@ibnesayeed I noticed that Reconstructive expects an absolute URI for the logo. Is there any way to surface the host/port info to this configuration? Ideally, a relative URI or one starting with a "/" (e.g., `/webui/images/ipwbLogo.png`) would be an easier to supply as a default.

ibnesayeed:
> I can reproduce it with the 5mementos WARC by going to the earliest memento, clicking the right arrow icon 6 times, and noting that the logo highlights.

I can fix that, but should we?

machawk1:
> but should we

[...]

PR #466, opened 2018-07-31
Link banner logo, prevent text selection, mute disabled nav links

ibnesayeed:
This implements most of the enlisted items in #465.


machawk1:
LGTM!

PR #468, opened 2018-08-07
Fixes #467 by checking for a CDXJ file's existence before performing …

### LINKED ISSUE #467: "Error" reported on viewport when searching in replay interface with no mementos
`ipwb replay non-existent.cdxj`

then access the replay UI at `localhost:5000`, noting `0 mementos` reported. Enter any query in the search/URI box and the resulting page shown simply says `Error`.

The terminal reports:
```
[Errno 2] No such file or directory: '/usr/local/lib/python2.7/site-packages/ipwb/usr/local/lib/python2.7/site-packages/ipwb/non-existent.cdxj'
(<type 'exceptions.IOError'>, IOError(2, 'No such file or directory'), <traceback object at 0x10b542bd8>)
```

machawk1:
…operations on it.

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -276,6 +276,11 @@ def getCDXJLinesWithURIR(urir, indexPath):
         indexPath = ipwbUtils.getIPWBReplayIndexPath()
     indexPath = getIndexFileFullPath(indexPath)
 
+    # Check that file exists prior to executing operations on it
+    if not os.path.isfile(indexPath):
+        print('The file at {0} does not exist.'.format(indexPath))
+        return []
+
     print('Getting CDXJ Lines with {0} in {1}'.format(urir, indexPath))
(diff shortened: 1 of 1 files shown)


ipwb/replay.py
```diff
@@ -276,6 +276,11 @@ def getCDXJLinesWithURIR(urir, indexPath):
         indexPath = ipwbUtils.getIPWBReplayIndexPath()
     indexPath = getIndexFileFullPath(indexPath)
 
+    # Check that file exists prior to executing operations on it
+    if not os.path.isfile(indexPath):
```
ibnesayeed: This check is performed in the wrong place. This function is being called later on various lookup requests. The existence of the index file should be checked at the time of [starting replay](https://github.com/oduwsdl/ipwb/blob/a6cecf1e9ae592e84120819a819ccf0beebd0bf6/ipwb/replay.py#L973) and should exit if the index is not found.

PR #469, opened 2018-08-07
Check for cdxj existence at startup per @ibnesayeed

machawk1:


ipwb/replay.py
```diff
@@ -992,6 +987,11 @@ def start(cdxjFilePath=INDEX_FILE, proxy=None):
         print('Sample data not pulled from IPFS.')
         print('Check that the IPFS daemon is running.')
 
+    # Check that supplied file exists
+    if not os.path.isfile(cdxjFilePath):
```
ibnesayeed: Now you are not white-listing `ipfs:` URIs. One one hand I am not a big fan of keeping the index file in IPFS, but on the other hand this check would fail the use-case you had implemented to allow an IPFS URI as index.

machawk1: This fails multiple tests as well.

ibnesayeed: I would instead encourage to figure out why tests are failing. There might be some assumptions made in the test file itself that is breaking here.

ibnesayeed: I think I know what went wrong here. The following test helper function is calling the `replay.start` function on line 44, but at that point the index file `tempFilePath` is only defined and not created. It will be created on line 51. This means the `start` function of the replay does not find the index file, hence exists. A quick fix would be to create an empty file at `tempFilePath` before `replay.start`.

https://github.com/oduwsdl/ipwb/blob/c2895694c0b17df3e8179e2fe7d64d180d62b0d4/tests/test_memento.py#L37-L52

ipwb/replay.py
```diff
@@ -975,6 +972,11 @@ def getCDXJLine_binarySearch(
         return lineFound
 
 
+def cdxjExists(pathOrHash):
```
ibnesayeed: This is currently a dead code. Also, if a local index file is supplied and does not exist, I think the process will not be killed with the current work flow of code.

machawk1: Oops, did not mean to leave that in there. It has been removed in 3806ea3.

machawk1: ...I just routed `start()` into existing logic and prevented an empty string from being returned in a scenario where we want to exit and inform the person running the script.

ibnesayeed: Perhaps we can create a separate ticket for the IPFS scheme and merge this one for now.

PR #491, opened 2018-08-10
Show ipwb replay CLI help when no parameters passed and not piping

### LINKED ISSUE #478: ipwb first-run is broken with default data

machawk1:
@ibnesayeed I realized I could pass the replayParser's help function in kwargs then execute it in the place we previously simply printed an error then returned. Please have a look at the code.

 Fixes #478

### DIFF
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -65,6 +65,7 @@ def checkArgs_replay(args):
         print('> ipwb replay /path/to/your/index.cdxj\n')
 
         args.onError()
+        sys.exit()
 
 
 def checkArgs(argsIn):
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -61,9 +61,10 @@ def checkArgs_replay(args):
     if suppliedIndexParameter:
         replay.start(cdxjFilePath=args.index, proxy=proxy)
     else:
-        # TODO: call replayParser.print_help()
-        print('Please supply a CDXJ file as an argument.')
-        sys.exit()
+        print('ERROR: An index file must be specified if not piping, e.g.,')
+        print('> ipwb replay /path/to/your/index.cdxj\n')
+
+        args.onError()
 
 
 def checkArgs(argsIn):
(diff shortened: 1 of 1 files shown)


ipwb/__main__.py
```diff
@@ -61,9 +61,10 @@ def checkArgs_replay(args):
     if suppliedIndexParameter:
         replay.start(cdxjFilePath=args.index, proxy=proxy)
     else:
-        # TODO: call replayParser.print_help()
-        print('Please supply a CDXJ file as an argument.')
-        sys.exit()
+        print('ERROR: An index file must be specified if not piping, e.g.,')
+        print('> ipwb replay /path/to/your/index.cdxj\n')
+
+        args.onError()
```
ibnesayeed: Don't we also need `sys.exit()` after calling this `onError()` function?

machawk1: @ibnesayeed No. When we had it before it was unnecessary. The program exits after the help function is called.

ibnesayeed: In that case we have many more places where `print_help` is being called and we have an explicit exit after that.

ibnesayeed: I can confirm that `print_help` does not exit automatically. If it is behaving like that here, the reason might be something else further down the execution chain. Here is a sample test script I wrote to verify it.

```python
import argparse
parser = argparse.ArgumentParser()
parser.parse_args()
print("BEFORE HELP")
parser.print_help()
print("AFTER HELP")
```

If `print_help()` were to cause a forced exit then we will not see the `AFTER HELP` message, but that's not the case.

```
$ python help_test.py
BEFORE HELP
usage: main.py [-h]

optional arguments:
  -h, --help  show this help message and exit
AFTER HELP
```

machawk1: @ibnesayeed Per my comment above, I think the reason for the exit is that there is no further execution in the callee. I believe the exiting to be independent of the `print_help` invocation and simply that the callee's code is complete.

ibnesayeed: Now I can see that the program terminates without an explicit exit here as nothing else is left to execute when the else branch is followed. This means we need an explicit `sys.exit()` here to ensure that the program behaves properly in situations where multiple commands are executed one after the other or an external program is watching for failures. Currently, the system would assume that everything is alright, though it is not.

machawk1:
@ibnesayeed I added an explicit exit following the printing of the replay help. Please let me know of any other changes need to accept this PR.

ibnesayeed:
The PR can be merged, but I think we need to fix one more thing about exits everywhere. We need to carefully check each instance of `sys.exit()` and specify a non-zero exit code in cases where the exit should suggest an error. By default the exit code is zero, which means everything is alright and program terminated normally.

Issue #473, opened 2018-08-07
Create a new endpoint to return a matched CDXJ record


ibnesayeed:
We need a new endpoint to return an index record instead of a reconstructed memento. This will enable us to try fetching IPFS blocks directly from the SW and reconstruct the memento there instead of letting the server do this. This will be a step in the direction of server-free decentralized replay. This will eliminate the need of threading as we can leverage asynchronous nature of JS for concurrent fetches. Additionally, we can avoid the location header rewriting issue (as per #456 and #461) by reusing the logic already present in Reconstructive.

`/cdxj/:datetime/:urir` should return 404 if no record is found, but return 200 (not a 3xx) otherwise with the one entry extracted from the index. We can either return `application/cdxj+ors` content type or `application/json` if we transform the index record into JSON.

machawk1:
@ibnesayeed This seems like the goal you described for server-less replay it would make achieving #434 even more difficult. Please comment on this, as I am motivated of integrating Prefer support as is relevant to my (our) research.

ibnesayeed:
No, the two are independent things. We can continue implementing support for `Prefer` header for raw memento. However, with client-side IPFS fetch, we will not need either of the raw or rewritten mementos as we will be performing the composition on the client side directly. Those accessing the server without the SW in place would need to talk to the regular memento endpoint.

machawk1:
@ibnesayeed Do you think the CDXJ meta headers should be included in the response?

I am working on an implementation that leverages our currently existing functionality and want to be sure I route through the right functions so as to not have to duplicate functionality.

ibnesayeed:
> Do you think the CDXJ meta headers should be included in the response?

We can think about that later when we start consuming the response. We might just create a JSON object that has all the necessary bits from the matched record and any other necessary metadata in it.

machawk1:
> ...just create a JSON object...

Per our verbal discussion, please outline how you expect this JSON object to look, e.g., including all the Memento-esque relations. Just an example ought to get us moving in the right direction to make this endpoint more useable for the replay banner.

machawk1:
@ibnesayeed Please document here the alternative Prefer semantics you described to me verbally in lieu of having a CDXJ endpoint.

ibnesayeed:
I think we are looking for something like this:

```
$ curl -i -H "Prefer: return=minimal" "http://localhost:5000/memento/20140115101500/memento.us/"
HTTP/1.0 200 
Preference-Applied: return=minimal
[...]

Issue #213, opened 2017-06-29
Require authorization for accessing admin functionalities on replay interface, allow smooth local access


machawk1:
Per #212, administrative functionality should not be available to anyone who can access the replay interface. @ibnesayeed offered a suggestion of having an `/admin/` endpoint. That seems like a big hammer for the likely more conventional use can of accessing the local instance.

Figure out a way to check if the client is local to the instance. If so, provide no-auth access. Otherwise, allow authenticated access using an `/admin/` endpoint. Also provide additional information (e.g., ipfs version), which might be sensitive, within the admin interface.

Said interface can also be added in-place on the main page using an iframe.

machawk1:
Checking `IPWBREPLAY_IP` equality with localhost (e.g., 127.0.0.1, localhost, etc) from the replay python script seems like a hacky way to determine this condition. Ideally we could tell by the HTTP headers in the request to the replay webUI but as far as I can see, there is no way to detect that using Flask.

ibnesayeed:
A few points here:

1) I could certainly not recommend an IP-based whitelisting because it is not feasible in situations when process runs in containers or on a remote server, while it also forces very strong restriction of "from where one can access certain features rather than who can access them"
2) We must segregate administrative features from informational stuff
3) I think "HTTP Basic Authentication" would serve the purpose and implementing that would be very simple in Flask (default credentials could be `admin/admin`, but modifiable using a CLI argument and an environment variable when the replay server instance is run). This approach would not require a full-blown authentication system with a database, while authenticated session automatically last quite long from the same browser

ibnesayeed:
* [A minimal custom HTTP Basic Auth implementation](http://flask.pocoo.org/snippets/8/)
* [An extension just for HTTP Basic Auth](https://flask-basicauth.readthedocs.io/en/latest/)
* [A more generic HTTP Auth extension that includes Basic Auth](https://flask-httpauth.readthedocs.io/en/latest/)

machawk1:
A preliminary barebones interface is now available at `/admin/` per ef26e2c4807b08b7a9a11de0382f3abc6d220fa1. Functionality from the main page is now to be filled in, styling applied, etc. and then it can be linked from the main replay page and the presence of admin functionality there removed.

PR #494, opened 2018-08-10
WARC Upload from replay UI

machawk1:


machawk1:
Something a little more elegant/intuitive should be done in the UI prior to merging.

Perhaps the ability to add things to the replay system via upload should be limited to admins but I can foresee a use case of trying to facilitate collaboration via uploading WARCs.

The UI elements ought to not be so prominent by default as well, since it's a side-feature. Maybe an "Upload WARC" button to toggle the expanded interface's display. There may be additional options we want to provide relating to the uploaded WARC in the future (e.g., specifying collection metadata), so hiding these now-few elements by default is likely the right way to go.

Though we anticipate drag-and-drop functionality, having this UI handy as well will make the feature more accessible.

@ibnesayeed We were supposed to talk about this in-person but I would like to know your opinion on the initial UI.

screen shot 2018-08-11 at 12 36 55 pm

ibnesayeed:
> Perhaps the ability to add things to the replay system via upload should be limited to admins but I can foresee a use case of trying to facilitate collaboration via uploading WARCs.

This is an interesting use case, but I can see a mechanism to do this without opening it up to everyone. In the admit interface we can add a temporary token (save in the memory as it is really temporary) button. This can generate one or more random long tokens that can then be used to temporarily authorize anyone to upload using a URL like `/uploads?access-token=<random-token>`. This way, an admin can invite people to contribute and then remove the token.

ibnesayeed:
> The UI elements ought to not be so prominent by default as well, since it's a side-feature. Maybe an "Upload WARC" button to toggle the expanded interface's display. There may be additional options we want to provide relating to the uploaded WARC in the future (e.g., specifying collection metadata), so hiding these now-few elements by default is likely the right way to go.

The WARC upload controls can go under the Options and Details section and be hidden by default. When that drawer is open then these controls would show up.

machawk1:
@ibnesayeed I like your idea for tokenization to allow users to upload.

However, I was thinking that the WARC upload interface would be independent of the Options and Details display, as that is hidden by default (we may want to encourage WARC uploads) but mainly, we plan to move the bulk of the admin functions in options and details to a separate interface (#213).

ibnesayeed:
The rationale why I think it should go inside of the drawer because in is an admin functionality and users of the replay should be bothered with that. Uploads are occasional and access is repeated, so surfacing non-frequent tasks up causes unnecessary clutter in the UI.

machawk1:
@ibnesayeed Ok. We should discuss the WARC Upload interface in-person. ;)

machawk1:
I think we also ought to consider giving the user the option to use the new CDXJ or revert to the one originally used to start the replay system. Maybe add a meta-endpoint to make available various CDXJs instead of simply the closest CDXJ resolution scheme as proposed in #473.

ibnesayeed:
That would require scoping which falls into the collection management territory (for which we have an existing ticket) and we will need to support index directories/files instead of a single index file in the replay.

machawk1:
@ibnesayeed WARC upload interface moved into the "drawer" per your suggestion.
screen shot 2018-08-14 at 6 02 12 pm

machawk1:
I believe the sole comment that needs to be addressed is to relocate the indexing logic back into the indexer and out of util, to where it was moved within this PR. Anything other requirements to get this merged, @ibnesayeed ?

ibnesayeed:
I have not tested it functionally yet, but on an architectural level that's the primary concern for now. Once the code is relocated and refactored as necessary, I will have another look and carefully follow the code flow.

ipwb/indexer.py
```diff
@@ -170,6 +170,83 @@ def indexFileAt(warcPaths, encryptionKey=None,
         print('\n'.join(cdxjLines))
 
 
+def generateCDXJMetadata(cdxjLines=None):
+    metadata = ['!context ["http://tools.ietf.org/html/rfc7089"]']
```
ibnesayeed: I remember mentioning somewhere that we need to change this context from the RFC to a separate URL (was that comment on your JCDL submission draft?).

ipwb/indexer.py
```diff
@@ -170,6 +170,83 @@ def indexFileAt(warcPaths, encryptionKey=None,
...
+        'generator': "InterPlanetary Wayback v.{0}".format(ipwbVersion),
+        'created_at': '{0}'.format(datetime.datetime.now().isoformat())
+    }
+    metaVals = '!meta {0}'.format(json.dumps(metaVals))
+    metadata.append(metaVals)
+
+    return metadata
+
+
+def joinCDXJFiles(cdxjPath1, cdxjPath2, outputFilePath):
```
ibnesayeed: Do we really need this new `joinCDXJFiles()` function? Can't we just call existing `indexFileAt()` function instead?

ipwb/indexer.py
```diff
@@ -170,6 +170,83 @@ def indexFileAt(warcPaths, encryptionKey=None,
...
+        cdxjMetadataLines = generateCDXJMetadata(cdxjLines)
+        cdxjLines = cdxjMetadataLines + cdxjLines
+
+        cdxjLines = '\n'.join(cdxjLines)
+
+    with open(outputFilePath, 'w') as wfd:
+        wfd.write(cdxjLines)
+
+
+def isCDXJMetadataRecord(cdxjLine):
```
ibnesayeed: This function can be as simple as:

```py
    return bool(cdxjLine.strip()) and cdxjLine.strip().startswith('!')
```

ipwb/indexer.py
machawk1: `indexFileAt` uses a WARC to generate a CDXJ, so no. `joinCDXJFiles()` takes two CDXJs as sources and merges them.

ipwb/indexer.py
machawk1: That refactoring can been performed after the PR. This function existed prior to this PR and is not original code. Please focus on the changes in this PR.

ipwb/indexer.py
machawk1: Perhaps but per below, this code is not new to this PR.

ipwb/indexer.py
ibnesayeed: In fact there can be an event shorter syntax:

```py
    return cdxjLine.strip()[:1] != '!'
```

ipwb/indexer.py
ibnesayeed: I think `joinCDXJFiles()` is redundant here, the `indexFileAt()` already performs merging when an outfile is specified and the outfile already exists. The function first reads whatever is in the outfile before indexing any new content. Then later truncates the file and writes the aggregated output to it. I remember implementing this functionality back when I added the outfile support.

machawk1: `indexFileAt` expects a WARC path as a parameter. It also considers encryption and other aspects. We need a function that, given two CDXJ files, merge them. Perhaps the logic in `indexFileAt()` can reference this new function.

ipwb/indexer.py
machawk1: I agree but again, this function is not introduced in this PR.

ipwb/indexer.py
ibnesayeed: No, we do not need to create a temporary incremental CDXJ file and a temporary combine CDXJ file and then reassign the index file to be used by the replay and piling up the throw away old CDXJ file and the incremental CDXJ file. A lot of this code can away by simply reusing what was already there. I will comment in the replay file where we need to make changes.

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+        filename = secure_filename(file.filename)
+        warcPath = os.path.join(app.config['UPLOAD_FOLDER'], filename)
+        file.save(warcPath)
+
+        random.seed()
+
+        cdxjPath = '/tmp/' + ''.join(random.sample(
+            string.ascii_uppercase + string.digits * 6, 12)) + '.cdxj'
+        combinedcdxjPath = '/tmp/' + ''.join(random.sample(
+            string.ascii_uppercase + string.digits * 6, 12)) + '.cdxj'
```
ibnesayeed: Rm both `cdxjPath` and `combinedcdxjPath` and perhaps `random.seed()` will not be needed here either.

Continue...

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+
+        # TODO: Check if semaphore lock exists, alert user if so, wait
+        # Index file, produce new.cdxj
+        print('Indexing file from uploaded WARC at {0} to {1}'.format(
+            warcPath, cdxjPath))
+        indexer.indexFileAt(warcPath, outfile=cdxjPath)
```
ibnesayeed: Use `app.cdxjFilePath` in place of `cdxjPath`.

Continue...

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+        print('index created at {0}'.format(cdxjPath))
```
ibnesayeed: This message should say, `Index updated at {app.cdxjFilePath}`

Continue...

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+
+        # TODO: Create semaphore lock
+        # Join current.cdxj w/ new.cdxj, write to combined.cdxj
+        print('* Prior index file: ' + app.cdxjFilePath)
+        print('* Index file of new WARC: ' + cdxjPath)
+        print('* Combined index file (to-write): ' + combinedcdxjPath)
```
ibnesayeed: These logs can be removed.

Continue...

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+        indexer.joinCDXJFiles(
+            app.cdxjFilePath, cdxjPath, combinedcdxjPath)
+        print('Setting ipwb replay index variables')
```
ibnesayeed: This `joinCDXJFiles()` call is not needed.

Continue...

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+
+        # Set replay.index to path of combined.cdxj
+        ipwbUtils.setIPWBReplayIndexPath(combinedcdxjPath)
+        app.cdxjFilePath = combinedcdxjPath
```
ibnesayeed: These `setIPWBReplayIndexPath()` and `cdxjFilePath` lines can go away as we are not changing the state of there.

Continue...

ipwb/replay.py
```diff
@@ -62,6 +74,60 @@ def setServerHeader(response):
...
+        app.cdxjFileContents = getIndexFileContents(combinedcdxjPath)
```
ibnesayeed: While we are not using `cdxjFileContents` anywhere, but keep it in sync, this should be updated using `app.cdxjFilePath` which by now has the updated/merged content.

ipwb/indexer.py
ibnesayeed: With the changes suggested in the replay and the removal of a big function in the indexer, the code is going to be a lot lighter, simpler to follow, easier to maintain, and functionally the same.

machawk1:
@ibnesayeed I made the changes you suggested in ef8a0ca291990fcc8b72187ae2371b6c6c369fdc

ibnesayeed:
I am looking at it and I find some issues, I am going to pull the code, revert you last commits and apply some changes then will push here.

ibnesayeed:
I didn't revert your changes, just restored a few things here and there and fixed missing pieces.

ibnesayeed:
Wait, there are a few other functions that can be reloacted for cleaner diffs.

machawk1:
Hmm, we don't use the variable `app.cdxjFileContents()` anywhere after setting it. What was the purpose of restoring it?

machawk1:
@ibnesayeed Go ahead and change them. If you would rather I do it, let me know.

ibnesayeed:
> Hmm, we don't use the variable app.cdxjFileContents() anywhere after setting it. What was the purpose of restoring it?

I restored because the original comment hinted about using some cache, I thought you might have plans on keeping the in memory and not reading the file each time. Which can be good for small indexes, but not so good for big ones on a limited memory machine. There is a refactoring needed in `getIndexFileContents()` as well, some of its functionalities belong to the startup process.

machawk1:
@ibnesayeed That was the original idea. Please refactor after getting this merged.

machawk1:
@ibnesayeed Something that might not show up in the tests but I just tested with 51e595d is that `app.cdxjFileContents = getIndexFileContents(combinedcdxjPath)` is still relying on the generated index that was being created when a WARC is uploaded that consists of the old index + new index combined.

machawk1:
Disregard, the instance was refactored out.

ibnesayeed:
> Disregard, the instance was refactored out.

I checked possible places before making that change, they were all expecting boolean value. Also, the change I made is a superset, but for existing cases it should behave exactly as it was before.

Issue #481, opened 2018-08-08
Add suggested BibTeX citation and rendered reference to README


machawk1:
The work @ibnesayeed and I verbally discussed to use as the BibTeX and rendered reference source was the TPDL 2016 publication, which is listed in the Project History section of the README.

PR #514, opened 2018-08-16
Change code highlight and prevent letter case of project name

ibnesayeed:
Improvement on #481.


ibnesayeed:
Actually, `PDF` is not a valid field on `inproceedings`, so we might want to delete that as some parsers might complain about it. However, we can provide the link to the paper separately for people to read.

machawk1:
I would be fine with that if you want to submit another PR (maybe also committing the PDF to the repo).

Issue #46, opened 2016-11-02
Conform to standardJS in webUI JavaScript


machawk1:
Related to #45 but relating to the code instead of meta for the repo.

https://github.com/feross/standard

machawk1:
Bump version req due to external project dependence (see #45)

machawk1:
webui is now up to date re:standardJS. However, `daemonController.js` calls a function in `webui.js`, so standard will fail on this. This needs to be remedied before this ticket is closed.

Additionally, we will exclude the Reconstructive JS files from being `standard`ized.

Issue #45, opened 2016-11-02
Add JS-standard linting for JavaScript in webUI (and elsewhere in project) to CI


machawk1:

machawk1:
It might not yet be possible to lint both JS and Python via TravisCI per https://github.com/travis-ci/travis-ci/issues/4090 . Given the project is 95% Python, I will keep with the pep8 validation atm.

machawk1:
Work moved to `pep8-standardJS-linting` branch

PR #518, opened 2018-08-17
Reconstructive and ReconstructiveBanner updates

ibnesayeed:
Updated Reconstructive and ReconstructiveBanner files to their latest version. Documentation comments are now removed and the banner has a more precise relative datetime along with an option to toggle between relative and absolute datetime display.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,12 @@ A pictorial representation of the ipwb indexing and replay process:
 
 ![image](https://raw.githubusercontent.com/oduwsdl/ipwb/master/docs/diagram_72.png)
 
+An important aspect of archival replay systems is rewriting various resource references for proper memento reconstruction so that they are dereferenced properly from the archive from around the same datetime as of the root memento and not from the live site (in which case the resource might have changed or gone missing). Many archival replay systems perform server-side rewriting, but it has its limitations when URIs are generated using JavaScript. To handle this we use [Service Worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) for rerouting requests on the client-side when they are dereferenced to avoid any server-side rewiring. For this, we have implemented a separate library, [Reconstructive](https://oduwsdl.github.io/Reconstructive/), which is reusable and extendable by any archival replay system.
+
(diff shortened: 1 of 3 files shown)


machawk1:
? cursor is not shown on hover of precise date.

ibnesayeed:
> ? cursor is not shown on hover of precise date.

That was intentional, because there is no surprise hidden when one goes back to the less precise representation,

machawk1:
If I toggle the display then visit the temporally adjacent memento, it reverts. This is frustrating and counter-intuitive.

ibnesayeed:
Again, you are talking about keeping the state over the session which is doable using something like storing state in local storage, but the essence of that option is different. Once a user is navigated to a new memento, it is an independent memento. Current philosophy is to show a gralnceable time difference with the option to uncover a precise datetime when demanded for it. Users' intent may or may not be to retain the state across navigation.

ibnesayeed:
While I do not disagree that there can be use niche cases where retaining this state might be useful for some users, at this point I do not see it as a preference, but an additional detail that user wanted to peek in (that is signified by the `?` cursor which indicates there is something more here). Some sites have keyword definitions or help tips associated with certain page elements which can be revealed when needed. If a user utilizes them and reloads to page or navigate to another page on the same site and they retain the open sate of those info boxes, it would be counter-intuitive. Similarly, if a site has a notification menu that can be be opened to see there were five new updates. The user clicks on one of those and has four more to go, then the site might not want to keep the notification menu open on each page load.

We expect that most of the users should open a memento to see what in the archive page. An archival banner is not the primary point of attraction on the page. It is there when it is needed otherwise it should go out of the way. With that in mind, I don't think it is bad to have the simplistic summary as the primary message and details on demand.

ibnesayeed:
Take another example, suppose a user is interested more details about a memento so uses banner in the expanded mode, while in this mode the user reloads the page, clicks on the nav buttons (next/prev/first/last), or enters a new URI-R to navigate to different memento, should we retain the banner in expanded view on the new page load just because the user might want to know more about that age as well? How would we know what is user's intent? Can't we default to what might work for most people and those who need it different can still spare one more click per page.

Internet archive has a drawer in it's banner that can be opened using `About this capture` link in the corner. But once you navigate to a new memento of the same URI-R or a different one, it goes away.

ibnesayeed:
> Aside: let's start complying with Standard JS (#45, #46), particularly re: superfluous semi-colons.

On the other hand I would suggest let's convert any non-Reconstructive code to the semi-colon style as those are minority in this repo. I am now using Babel, a very well-known JS transpiler which continues to use semi-colons by default and the whole MDN documentation uses this practice of including semi-colons in in the modern JS APIs.

ibnesayeed:
Changes in #520 took care of the state persistence feature and more.

Issue #178, opened 2017-06-06
Pretty listing of resources and title for HTML pages


ibnesayeed:
Currently the bare URLs are listed for the archived resources (as shown in #177). We can make it better looking and less space consuming by:

* Showing titles of HTML pages and hyperlinking corresponding mementos
* For resources where title is not present or is not applicable, their file name can be extracted from the URL
* As a fallback, URLs can be used if nothing else is feasible
* Along with the title, if generated, thumbnails would also be a good way to present resources.

Title of each memento can be stored in the CDXJ file as an optional filed. Title extraction would require HTML parsing at the time of indexing.

machawk1:
@ibnesayeed Do you believe title extraction should be the default functionality or only activated when a flag is passed to the indexing script?

My vote is the former, though it makes the CDXJ more verbose but richer and more user-friendly when parsed and displayed. We may also offer the option to enrich CDXJ TimeMaps that do not have this information from within the replay interface, e.g., 

<details>
<summary>A sample CDXJ w/o title attributes</summary>
!context ["http://oduwsdl.github.io/contexts/cdxj"]
!meta {"created_at": "2017-07-14T09:02:23.458675", "generator": "InterPlanetary Wayback v.0.2017.07.10.1739"}
com,matkelly)/froggies/frog.png 20170301192639 {"locator": "urn:ipfs/QmUeko8zM7Xanwz6F9GtRH4rLAi4Poj3DMECGsci2BRQfs/QmPhMnX74cwqx2xgj9d3N3gTra8CzafXwSbUwU8xagMfqR", "mime_type": "image/png", "status_code": "200"}
com,matkelly)/robots.txt 20170301192639 {"locator": "urn:ipfs/Qmbk3Aju7u26Pzk356a43wY9eUCScAJiLPxhvwsMoVt7Pd/QmYNB85U2txRAAdLp6wvZSPvd8AQq8UcjZJ2azhv5h6NF7", "mime_type": "text/plain", "status_code": "200"}
edu,odu,cs)/~mkelly/semester/2017_spring/remotefroggie.html 20170301192639 {"locator": "urn:ipfs/QmPdyY6Pm66iWtGpTc7PqK11hvsnYSKMVL57G69RiNjGcm/QmNZ6mKSSAXAmXEocQj5gT4y4kdcr5D2C173ubWJ6PSKEZ", "mime_type": "text/html", "status_code": "200"}
</details>

...
[...]

PR #527, opened 2018-08-26
Extract title from payload and add to the index when available

ibnesayeed:
Related to #178

### DIFF
--- a/ipwb/indexer.py
+++ b/ipwb/indexer.py
@@ -229,7 +229,7 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
             title = None
             try:
                 ctype = hdrs.get_header('Content-Type')
-                if (ctype and ctype.lower().startswith('text/html')):
+                if ctype and ctype.lower().startswith('text/html'):
                     title = BeautifulSoup(payload, 'html.parser').title.text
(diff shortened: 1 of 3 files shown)


ipwb/indexer.py
```diff
@@ -224,6 +226,16 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
             entry.buffer.seek(0)
             payload = entry.buffer.read()
 
+            title = None
+            try:
+                ctype = hdrs.get_header('Content-Type')
+                if (ctype and ctype.lower().startswith('text/html')):
```
machawk1: Are the outer parens necessary? We don't use that form elsewhere.

machawk1:
LGTM aside from the above outer parens question. Please answer that, @ibnesayeed, then we can merge this into master.

ipwb/indexer.py
ibnesayeed: No, these are not necessary. Initially, I was using header getter inline and did not have that `ctype` variable, so the conditional line was exceeding the pycodestyle limits. In that case generally parens are placed and lines are broken. later, I extracted the repeated part in a variable and it was not needed anymore, but forgot to remove it. Will send commit soon to fix it.

PR #533, opened 2018-08-27
Check if a title exists before trying to extract it. Closes #532

### LINKED ISSUE #532: Title extraction when indexing a page has no title produces error
```
ipwb index ipwb/samples/warcs/5mementos.warc | ipwb replay --proxy=localhost:5002
Failed to extract title in 5mementos.warc: 0/7
'NoneType' object has no attribute 'text'
...
```
There ought to be a check in the indexer that `title = BeautifulSoup(payload, 'html.parser').title` is not `None` before accessing the `title = BeautifulSoup(payload, 'html.parser').title.text` attribute.

machawk1:


ipwb/indexer.py
```diff
@@ -230,8 +230,9 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
             try:
                 ctype = hdrs.get_header('Content-Type')
                 if ctype and ctype.lower().startswith('text/html'):
-                    title = BeautifulSoup(payload, 'html.parser').title.text
-                    title = ' '.join(title.split()) or None
+                    title = BeautifulSoup(payload, 'html.parser').title
+                    if title is not None:
+                        title = ' '.join(title.text.split())
```
ibnesayeed: Please change this line to `title = ' '.join(title.split()) or None` otherwise there will be cases where it will return an empty string. Doing this will make sure either we have a non-empty title or return `None`.

machawk1: @ibnesayeed Done in 95aba90. It was unclear why you put the `or None` logic, hence the removal.

Issue #382, opened 2018-03-13
Licensing issue of ipwb's MIT w/ imported GPL libs


machawk1:
Stemming from the discussion in https://github.com/datatogether/roadmap/issues/79, ipwb is MIT but uses some https://github.com/webrecorder/pywb modules, the latter being GPLv3.

There have been numerous discussions in this repo of decoupling ipwb from pywb, so that horse need not be beaten any further, but is there any issue of an MIT package using a GPLv3 package?

/cc @ibnesayeed

machawk1:
> However, this issue can easily be resolved by utilizing warcio which is released under MIT-compatible Apache 2.0 license.

That would solve the issue for the WARC-related functions we use from pywb but there is other reuse, e.g., https://github.com/oduwsdl/ipwb/blob/68e8e1d5ea59585d5a8ee599e86834bdba3687c2/ipwb/replay.py#L24-L25

Since it's "easily be resolved", can I assume that you are going to "easily resolve it"? ;)

#380 is an outstanding related issue.

ibnesayeed:
I can see that the binary search function need to be implemented as well, but that would be a bitesize task that could be a good way to engage some new WS-DL members in this project.

The [SURT implementation](https://github.com/rajbot/surt) exists outside IPWB, but unfortunately that is under an AGPL-3.0 license. Our best bet would be implement a simple SURT library and release it under MIT. That could be another good task for newcomers while we can elaborate on how it can be done.

> Since it's "easily be resolved", can I assume that you are going to "easily resolve it"? ;)

#AfterABD ;-)

machawk1:
For the most part, we have moved to using the Apache 2.0-licenses warcio instead of importing parts of the pywb module. An outstanding issue that prevents this ticket from being closed is that [we have copied pywb's unsurt function](https://github.com/oduwsdl/ipwb/blob/95ec3c6231a33366f16b4b92be9dde32bf8dc609/ipwb/util.py#L254-L266) for use in ipwb directly without needing pywb as an explicit dependency. While credit is maintained, this seems like an approach that exacerbates the license incompatibility issues between ipwb (MIT) and pywb (GPLv3).

ibnesayeed:
Even the `surt` library is AGPL that we are still using.

machawk1:
Ok, then the problem is _worse_ than originally reported.

PR #535, opened 2018-08-27
[WIP] Fix plurality of mementos and resources on main replay page. Closes #534

### LINKED ISSUE #534: Plurality issue in the URI listing in the replay UI
screen shot 2018-08-27 at 11 57 17 am

It should be "1 memento of 1 resources with 0 HTML pages listed"

machawk1:
@ibnesayeed Kudos on the implementing the templating system, btw!

### DIFF
--- a/ipwb/assets/webui.js
+++ b/ipwb/assets/webui.js
@@ -131,12 +131,8 @@ function setShowURIsVisibility () {
 }
 
 function setPlurality () {
-  const urimCount = document.getElementById('memCountInt').innerHTML
   const htmlFilesPlurality = document.getElementById('htmlPages').innerHTML
 
-  if (urimCount === '1') {
-    document.getElementById('plural').classList.add('hidden')
-  }
   if (htmlFilesPlurality === '1') {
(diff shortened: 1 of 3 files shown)


machawk1:
@ibnesayeed That would be a good improvement but for now, this fixes the issue described in #534.

machawk1:
`inflect` is also GPLv3, @ibnesayeed (related: #382)

ibnesayeed:
I will leave the decision of merging this PR in its current state or otherwise to you discretion because I am not a big fan of hacky approaches on every step that pile up and make maintenance difficult (and invite more hacks for every new change).

ibnesayeed:
> inflect is also GPLv3

We can avoid using that, but still move the pluralization logic into Context Processors.

machawk1:
@ibnesayeed Ok, I'll get a little more familiar with Flask's context processors and implement that with this same PR.

machawk1:
`./ipwb/replay.py:73:5: E731 do not assign a lambda expression, use a def` needs to be fixed before we merge this.

PR #538, opened 2018-08-28
Rename the daemon and admin endpoint

machawk1:
Per in-person discussion

### DIFF
--- a/ipwb/assets/serviceWorker.js
+++ b/ipwb/assets/serviceWorker.js
@@ -29,7 +29,7 @@ const rc = new Reconstructive({
 // <   localResource: function(FetchEvent) => boolean
 // < }
 rc.exclusions.specialEndpint = function (event, config) {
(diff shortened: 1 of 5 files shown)


ipwb/assets/serviceWorker.js
```diff
@@ -29,7 +29,7 @@ const rc = new Reconstructive({
 // <   localResource: function(FetchEvent) => boolean
 // < }
 rc.exclusions.specialEndpint = function (event, config) {
-  return ['/ipwbassets/', '/daemon/', '/config/'].some(
+  return ['/ipwbassets/', '/ipfsdaemon/', '/config/'].some(
```
ibnesayeed: Should we also change `/config` to `/ipwbconfig`?

machawk1: `/config/` updated in 26436f9.

Issue #51, opened 2016-12-05
Migrate code from Python 2.7.x to Python 3


ibnesayeed:
@machawk1, is there anything that is stopping us from moving this to to Python 3.x?

machawk1:
@ibnesayeed I have not checked compatibility of all of the included libraries w/ Py3. What advantage would we gain in doing so? There are more fundamental problems w/ the project to be tackled than updating the code to the latest version of the language.

ibnesayeed:
At this point it is not a priority, but for the longevity of the project it should not use the version of the language that is nearing the extended EOL date.

machawk1:
Ok, it will be considered in the future when more specific rationales are put forth.

ibnesayeed:
```
$ git clone https://github.com/oduwsdl/ipwb
$ cd ipwb/
$ 2to3 -W .
$ pip install -r requirements.txt
$ pip install ./
$ ipwb
Traceback (most recent call last):
  File "/usr/local/bin/ipwb", line 9, in <module>
    load_entry_point('ipwb==0.2016.12.9.412', 'console_scripts', 'ipwb')()
  File "/usr/local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/usr/local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/usr/local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.5/site-packages/ipwb/__main__.py", line 6, in <module>
    from . import replay
  File "/usr/local/lib/python3.5/site-packages/ipwb/replay.py", line 19, in <module>
    from . import util as ipwbConfig
  File "/usr/local/lib/python3.5/site-packages/ipwb/util.py", line 7, in <module>
    import exceptions
ImportError: No module named 'exceptions'
```

machawk1:
Same commands from virtualenv:

```sh
ipwb
Traceback (most recent call last):
  File "/usr/local/bin/ipwb", line 11, in <module>
[...]

PR #547, opened 2018-08-31
Tweak version check to be inter-Py-compatible

machawk1:


machawk1:
Related to #51 (no `cmp()` in Py3)

PR #550, opened 2018-09-03
Add max and min dates available to admin interface

machawk1:


machawk1:
There is much room for improvement here but it ties in the logic of polling the CDXJ for information about the captures. Improvements include prettifying dates, logic to not show a range when there is only one memento, consolidation of commonality (e.g., June 3-6, 2018 instead of June 3, 2018 to June 6, 2018), etc.

PR #552, opened 2018-09-10
Mv sample files outside of package, consolidate test samples. Closes #540

### LINKED ISSUE #540: Do not package sample files
Since we no longer use a default warc file for replay, there is no incentive to package all the sample files which makes the package unnecessarily big.

Also, currently there are two places where samples are stored, one under the package and the other under the test directory. Some test files use resources from both. We can consolidate them in a single folder and move it up in the hierarchy (i.e., directly under the project root).

machawk1:
…#540


tests/testUtil.py
```diff
@@ -17,7 +17,7 @@
 def createUniqueWARC():
     lines = []
     warcInFilename = 'frogTest.warc'
-    warcInPath = os.path.join(os.path.dirname(__file__) + '/samples/warcs/' +
+    warcInPath = os.path.join('samples/warcs/' +
```
ibnesayeed: Are these relative paths stable? I feel like they will be resolved based on the context from where the tests are run and not the location of the file itself. This is just a guess, I have not verified it. A safer approach would be something like `os.path.dirname(__file__) + '/../samples/warcs/`.

machawk1: @ibnesayeed For pytest, yes, they should be stable. The context should be maintained for each test at the root of the package.

ibnesayeed: Allow me to test something.

machawk1: Test to your heart's content. This PR has already been merged.

ibnesayeed: As I guessed, it breaks miserably when tested from anywhere other than the root of the repository. While our automated test is set to run from that context, there is no reason why these references should not be made more robust. We might want to create a variable to hold the absolute path of the samples directory and use it in different places as necessary to construct more specific paths:

```py
samples_dir = os.path.join(os.path.dirname(__file__) + '/../samples/'
```

ibnesayeed: Opened a separate ticket (#553) for this.

PR #557, opened 2018-09-18
Issue 556

machawk1:


ipwb/assets/webui.js
```diff
@@ -18,22 +18,27 @@ function hideURIs () {
...
+      a.appendChild(document.createTextNode(urir))
+      dt = document.createTextNode(' (' + memento['datetime'] + ')')
 
       li.appendChild(a)
       li.appendChild(dt)
-      li.setAttribute('data-mime', uris[uri]['mime'])
+
+      li.setAttribute('data-mime', memento['mime'])
+      li.setAttribute('data-status', memento['status'])
+      if (memento['mime'].startsWith('text/html') && memento['status'][0] !== '3') {
```
ibnesayeed: How about?

```js
if (memento['mime'].toLowerCase().startsWith('text/html') && memento['status'][0] === '2') {
```

1. People might have upper-case content-type and
2. We might have recorded `5xx` being replayed (so it's better to explicitly check for `2xx` instead of lack of presence of `3xx`)

machawk1: Checking for 2XX will prevent archived 404s from being replayed. I will make the change regarding capitalization.

ibnesayeed: Do we want to list 4XXs and 5XXs?

machawk1: Yes. They might be interesting to study. If all-we-have is a 4XX, it is still a representation of the URI-M. As you are aware, @ibnesayeed, we will eventually not have this default comprehensive list. At that point, perhaps we can have some other indicator to distinguish 2XX, 3XX, 4XX, 5XX in the comprehensive list.

PR #563, opened 2018-09-19
Regain sort order in main replay URI list. Closes #562

### LINKED ISSUE #562: New dead code in JS
After new changes in the resource listing logic, following code is now useless (it is executed and not unreachable though).

https://github.com/oduwsdl/ipwb/blob/f0f7fdbee058b909ea39b8c44fc1af3252e92546/ipwb/assets/webui.js#L20

With that in effect, `shortestFirst` function is also useless now.

machawk1:


ibnesayeed:
By the way, that was an untested code suggestion, hope you have tested it before pushing here.

PR #566, opened 2018-09-19
Split date from time and position it before URI-R in listing

ibnesayeed:
This PR changes URI listing from:

* memento.us/ (20161231110001)
* archive.org/ (20080430204826)

to

* [2016-12-31 11:00:01] memento.us/
* [2008-04-30 20:48:26] archive.org/


ibnesayeed:
The primary reason of putting dates first was a good readable alignment (as you noted it). Also, the date is not linked, so it should not look as bad because the colored links will capture the gaze first.

machawk1:
@ibnesayeed I disagree. Maybe it's the implicit precedence of putting the date before the URI-R,  the switch to a longer (but more visually parse-able) date string (missing TZ, btw), or the square brackets over parens but I prefer the method displayed in current master over what this PR is offering despite its (mentioned) merits.

Please add a little more to this PR per above and #178 (title) for approval.

ibnesayeed:
Parens were good when the date time was rendered after the URL (and title in future), but if we want to place it up front, perhaps square brackets would make more sense.

With dates and times split with delimiters, placing it after the link looks really ugly when half of it breaks into a new line.

ibnesayeed:
### Non-wrapped entries

#### datetime after

ipwb-urilist-nowrap-date-after

#### datetime before

ipwb-urilist-nowrap-date-before

### Wrapped entries

#### datetime after

ipwb-urilist-date-after

#### datetime before

ipwb-urilist-date-before

I find the datetime before the URI more readable and well aligned in both the cases. That said, there is the possibility to force the datetime record never wrap, but it didn't look very good (to me) either.

machawk1:
I see another merit in date-first in that it will always be displayed in the case of a long URI. We ought to adjust the text-overflow property of the URIs. Let's still get the titles working. Let me sleep on this PR.

PR #570, opened 2018-09-25
Surfaces the titles to the webUI and associates with the respective JSON and links for #178

machawk1:
They are current associated with the `li`'s `title` DOM attribute. We can change this in the future but having the data associated with the memento listing is a first step toward #178.

### DIFF
--- a/ipwb/assets/webui.js
+++ b/ipwb/assets/webui.js
@@ -32,6 +32,9 @@ function addURIListToDOM () {
 
       li.setAttribute('data-mime', memento['mime'])
       li.setAttribute('data-status', memento['status'])
+      if ('title' in memento) {
+        li.setAttribute('title', memento['title'])
+      }
 
       const isHTML = memento['mime'].toLowerCase().startsWith('text/html')
       const isARedirect = memento['status'][0] === '3'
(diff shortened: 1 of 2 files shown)


ipwb/assets/webui.js
```diff
@@ -32,6 +32,9 @@ function addURIListToDOM () {
 
       li.setAttribute('data-mime', memento['mime'])
       li.setAttribute('data-status', memento['status'])
+      if ('title' in memento) {
+        li.setAttribute('title', memento['title'])
```
ibnesayeed: If we are planning to merge #566 (after title inclusion) then this change in the `webui.js` is not desired as it will cause unnecessary conflict. Making the title data available for usage should be good enough for this PR.

machawk1: @ibnesayeed This association is an effort to get #566 merged. It currently does not deal with titles. By associating it in the UI with the `li`, we can work toward including it in the listing.

ibnesayeed: @machawk1 #566 will be able to directly use the `title` attribute from the `memento` object. We don't need to add it to each `li` beforehand.

machawk1: @ibnesayeed Rm'd in 8b866e6. Once this PR is merged, we can prettify #566 a bit more before merging it in.

Issue #485, opened 2018-08-09
0 length response bodies cause IPFS add to fail


machawk1:
This occurs in WARCs with WARC Response records containing HTTP responses with a 204 No Content status code.

When `pushToIPFS()` in the indexer calls `pushBytesToIPFS()`, an `ipfsapi.exceptions.ConnectionError` exception occurs.

The start to a fix would be to check these values' length upon extraction.

machawk1:
Replication:
```python
> import ipfsapi
> IPFS_API = ipfsapi.Client('localhost', 5001)
> IPFS_API.add_bytes('lorem')
u'QmeHxdVgh5BVLL7S7AvjrXqum8977XcPcfyHYUkoCG624N'
> IPFS_API.add_bytes('')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/ipfsapi/utils.py", line 150, in wrapper
    res = cmd(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/ipfsapi/client.py", line 2087, in add_bytes
    data=body, headers=headers, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/ipfsapi/http.py", line 37, in wrapper
    return func(self, *args, **merged)
  File "/usr/local/lib/python2.7/site-packages/ipfsapi/http.py", line 170, in request
    files, headers, data)
  File "/usr/local/lib/python2.7/site-packages/ipfsapi/http.py", line 100, in _request
    files=files, headers=headers, data=data)
  File "/usr/local/lib/python2.7/site-packages/ipfsapi/http.py", line 77, in _do_request
    six.raise_from(exceptions.ConnectionError(error), error)
  File "/Users/machawk1/Library/Python/2.7/lib/python/site-packages/six.py", line 737, in raise_from
    raise value
ipfsapi.exceptions.ConnectionError: ConnectionError: [Errno 32] Broken pipe
>>> 
```

machawk1:
This arose in the WARC provided by @james-e-powell as generated by wget, which encountered a 204 No Content when archiving a GitHub status page.

machawk1:
Also reported https://github.com/ipfs/py-ipfs-api/issues/137 but was reminded that py-ipfs-api has deprecated Python 2 usage (I think...the README line about the deprecated is ambiguous).

machawk1:
[...]

PR #572, opened 2018-09-25
Mitigate noise from py-ipfs-api due to its mis-handling of 0-length payloads per #485

machawk1:
The issue of using py-ipfs-api to `add_bytes()` of zero-length payloads is still an outstanding issue per https://github.com/ipfs/py-ipfs-api/issues/137. This causes a _lot_ of noise by the ipwb indexer when encountering these records, like when indexing `samples/warcs/IAH-20080430204825-00000-blackbook.warc.gz`.

This PR prevents these records from being _attempted_ to be pushed into IPFS using py-ipfs-api. When https://github.com/ipfs/py-ipfs-api/issues/137 is fixed, we can change the logic but for now, this will short-circuit the attempt from ipwb.

### DIFF
--- a/ipwb/indexer.py
+++ b/ipwb/indexer.py
@@ -71,6 +71,9 @@ def pushToIPFS(hstr, payload):
             if isinstance(payload, str):
                 payload = s2b(payload)
 
+            if len(payload) == 0:  # py-ipfs-api issue #137
+                return
+
             httpHeaderIPFSHash = pushBytesToIPFS(hstr)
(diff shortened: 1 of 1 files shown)


ibnesayeed:
We might want to add a special hash for now (e.g., `0000000000000000000...`) to indicate zero bytes. And then check this well known hash in the replay to not attempt to fetch it. Current changes will ignore such records from being replayed as they are skipped from index.

machawk1:
Merging this for now. We can revisit it later but keeping the status quo makes for messy testing.

PR #573, opened 2018-09-25
Considers the XHTML MIME type when displaying 'HTML pages' in replay …

### LINKED ISSUE #558: XHTML pages are not considered HTML and are not displayed by default in replay
Per #556, mementos of the MIME type `application/xhtml+xml` are not displayed by default on the main replay page despite being "HTML".

machawk1:
…interface. Closes #558 

Tested with 2mementos_htmlXhtml.warc in master sample WARCs.

### DIFF
--- a/ipwb/assets/webui.js
+++ b/ipwb/assets/webui.js
@@ -39,7 +39,10 @@ function addURIListToDOM () {
       li.setAttribute('data-mime', memento['mime'])
       li.setAttribute('data-status', memento['status'])
 
-      const isHTML = memento['mime'].toLowerCase().startsWith('text/html')
+      const mementoMIME = memento['mime'].toLowerCase()
+      const isHTML = mementoMIME.startsWith('text/html') ||
+        mementoMIME.startsWith('application/xhtml+xml')
+
       const isARedirect = memento['status'][0] === '3'
       if (isHTML && !isARedirect) {
         li.setAttribute('data-display', 'default')


ipwb/assets/webui.js
```diff
@@ -39,7 +39,10 @@ function addURIListToDOM () {
       li.setAttribute('data-mime', memento['mime'])
       li.setAttribute('data-status', memento['status'])
 
-      const isHTML = memento['mime'].toLowerCase().startsWith('text/html')
+      const mementoMIME = memento['mime'].toLowerCase()
+      const isHTML = mementoMIME.startsWith('text/html') ||
+        mementoMIME.startsWith('application/xhtml+xml')
```
ibnesayeed: How about something like which will be easier to add more items in the future:

```js
      const htmlMIMEs = ['text/html', 'application/xhtml+xml']
      const mementoMIME = memento['mime'].split(' ')[0].toLowerCase()
      const isHTML = htmlMIMEs.includes(mementoMIME)
```

machawk1: @ibnesayeed I think this would lose our `startsWith()` logic where a MIME type of `nottext/html` would match your revision logic but not the one in this PR.

Also, LGTM _and_ an ❌ edit!? 🤔

ibnesayeed: No, it will not lose that ability to tight match.

```js
['text/html', 'application/xhtml+xml'].includes('nottext/html')
> false
```

I didn't mean LGTM, It was perhaps somewhere in the buffer of the form as I might have typed earlier.

machawk1: @ibnesayeed A space need not be mandatory in the Content-type, e.g., `text/html;charset=utf-8` per https://tools.ietf.org/html/rfc7231#section-3.1.1.5. Your code snippet would fail with this despite it being an HTML MIME type.

PR #576, opened 2018-09-25
Return empty MIME in JSON instead of null

### LINKED ISSUE #575: Return empty MIME in JSON instead of null
When MIME is `None`, return an empty string in JSON.

ibnesayeed:
This should fix #575.

### DIFF
--- a/ipwb/indexer.py
+++ b/ipwb/indexer.py
@@ -287,7 +287,7 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
                 'locator': 'urn:ipfs/{0}/{1}'.format(
                     httpHeaderIPFSHash, payloadIPFSHash),
                 'status_code': statusCode,
-                'mime_type': mime,
+                'mime_type': mime or '',
                 'original_uri': originaluri
             }
             if encCompOpts.get('encryptionKey') is not None:
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -944,7 +944,7 @@ def getURIsAndDatetimesInCDXJ(cdxjFilePath=INDEX_FILE):
 
         mementoAsJSON = {
             'datetime': datetime,
-            'mime': jsonFields['mime_type'],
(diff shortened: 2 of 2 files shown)


machawk1:
@ibnesayeed It seems that the logic directly below your edit for the title could use that `or` logic as well. Thoughts?

ibnesayeed:
> It seems that the logic directly below your edit for the title could use that or logic as well. Thoughts?

It can, but the two fields are represented a little differently. Since MIME is an almost essential field, we always provide it in the JSON response (even if it is empty from some resources), but the title will only be present in a small subset of resources, so we do not include that in the JSON if it is not present.

machawk1:
@ibnesayeed ok. Can you give your PR some testing? If not, I will do so when I get the chance.

ibnesayeed:
I have tested it and it seems to be working as expected.

machawk1:
@ibnesayeed Do you think perhaps the indexer ought not put put the mime as null in the index?

In dae45da7793e23e736c281e54fcbd112c327c52e:
```
# ipwb index samples/warcs/1memento_noContentType.warc 
Processing WARC records in 1memento_noContentType.warc complete
!context ["http://tools.ietf.org/html/rfc7089"]
!meta {"created_at": "2018-09-25T21:08:11.755723", "generator": "InterPlanetary Wayback v.0.2018.09.25.1619"}
us,memento)/ 20130202100000 {"locator": "urn:ipfs/QmRDR4e6JxddXt5XtC73WUV6D6TCBr7z7QRh48vqjATaid/QmZio6GwXQ3o1XjGchsbUBAMgLWifW5L44XbzT95C5PyyH", "original_uri": "http://memento.us/", "mime_type": null, "status_code": "200"}
```

I think nipping the unspecified value in the bud is better than trying to mitigate the value later.

ibnesayeed:
I think we should suppress that `null` value from the index and place an empty string instead.

Issue #579, opened 2018-10-01
TimeGate does not respond correctly to malformed datetimes


machawk1:
Section 4.5.3 of the [Memento RFC](https://tools.ietf.org/html/rfc7089) says

> If the value of the "Accept-Datetime" does not conform to the rfc1123-date construction rule of the BNF in Figure 1, the response MUST have a "400 Bad Request" HTTP status code.

This is not exhibited in the ipwb replay system.

This issue was discovered using the [Deployed ipwb instance](https://ipwb.ws-dl.cs.odu.edu/) and the [Memento Validator Service](http://mementoweb.org/tools/validator/) with https://ipwb.ws-dl.cs.odu.edu/timegate/memento.us as the URI and TimeGate as the Type.

machawk1:
Closed in #580

PR #580, opened 2018-10-01
Return HTTP 400 when Accept-Datetime is not RFC1123 + tests

machawk1:


machawk1:
This would close #579

ipwb/util.py
```diff
@@ -146,6 +146,14 @@ def iso8601ToDigits14(iso8601DateString):
     return d.strftime('%Y%m%d%H%M%S')
 
 
+def isRFC1123Compliant(dtstr):
```
ibnesayeed: @machawk1 I think this method is wasteful, duplicates the code, and causes the same logic to be executed twice. This same effect could have been achieved by catching the `ValueError` exception in the `replay.py` file when calling `ipwbUtils.rfc1123ToDigits14(adt)` function.

ipwb/replay.py
```diff
@@ -315,6 +315,9 @@ def queryTimeGate(urir):
     if adt is None:
         adt = ipwbUtils.getRFC1123OfNow()
 
+    if not ipwbUtils.isRFC1123Compliant(adt):
+        abort(400)
+
```
ibnesayeed: No need to introduce `isRFC1123Compliant` function. We can wrap the following line in a `try` block to catch `ValueError` exception instead.

ipwb/util.py
machawk1: @ibnesayeed I anticipate this function to be useful elsewhere, hence the inclusion in util and not hard-coded to the use case in replay.

ipwb/replay.py
machawk1: See response to other comment.

ipwb/util.py
ibnesayeed: To deal with parsing errors like this, exceptions are generally a better way, no matter how often you use them.

PR #584, opened 2018-10-05
Prevent Flask/werkzeug from improperly converting relative location r…

### LINKED ISSUE #582: TimeGate does not inherit scheme for redirects
In [validating](http://mementoweb.org/tools/validator/info.html?uri=https%3A%2F%2Fipwb.ws-dl.cs.odu.edu%2Ftimegate%2Fmemento.us%2F&type=timegate&datetime=Sun%2C+01+Apr+2010+12%3A00%3A00+GMT&aggregator=http%3A%2F%2Ftimetravel.mementoweb.org%2Ftimegate%2F&follow=on&Submit=Submit) the ipwb Memento compliance, I noticed the warning of

> TimeGate does not redirect to last memento for datetimes in the future

This exact match check may be due to the difference in schemes (http vs https) that the TimeGate endpoint is serving in the location header.

```
$ curl -i -H "Accept-Datetime: Thu, 31 May 2099 20:35:00 GMT" https://ipwb.ws-dl.cs.odu.edu/timegate/memento.us/
HTTP/2 302 
content-type: text/html; charset=utf-8
date: Tue, 02 Oct 2018 14:46:37 GMT
link: <http://memento.us/>; rel="original", <https://ipwb.ws-dl.cs.odu.edu/timemap/link/memento.us/>; rel="timemap"; type="application/link-format", <https://ipwb.ws-dl.cs.odu.edu/timemap/cdxj/memento.us/>; rel="timemap"; type="application/cdxj+ors", <https://ipwb.ws-dl.cs.odu.edu/timegate/memento.us/>; rel="timegate", <https://ipwb.ws-dl.cs.odu.edu/memento/20130202100000/memento.us/>; rel="first memento"; datetime="Sat, 02 Feb 2013 10:00:00 GMT", <https://ipwb.ws-dl.cs.odu.edu/memento/20161231110000/memento.us/>; rel="prev memento"; datetime="Sat, 31 Dec 2016 11:00:00 GMT", <https://ipwb.ws-dl.cs.odu.edu/memento/20161231110001/memento.us/>; rel="last memento"; datetime="Sat, 31 Dec 2016 11:00:01 GMT"
location: http://ipwb.ws-dl.cs.odu.edu/memento/20161231110001/memento.us/
server: InterPlanetary Wayback Replay/0.2018.10.01.1947
vary: Accept-Datetime
content-length: 277

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>Redirecting...</title>
<h1>Redirecting...</h1>
<p>You should be redirected automatically to target URL: <a href="/memento/20161231110001/memento.us/">/memento/20161231110001/memento.us/</a>.  If not click the link.
```

Note the difference in scheme of the `last` entry in the Link headers and the value in the `location` header:
```
https://ipwb.ws-dl.cs.odu.edu/memento/20161231110001/memento.us/
http://ipwb.ws-dl.cs.odu.edu/memento/20161231110001/memento.us/
```

This could be due to some hard-coding in the replay system, to be investigated.

machawk1:
…edirects to absolute. Closes #582

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -89,16 +89,19 @@ def allowed_file(filename):
 @app.route('/upload', methods=['POST'])
 def upload_file():
     # check if the post request has the file part
+    resp = redirect(request.url)
+    resp.autocorrect_location_header = False
+
     if 'file' not in request.files:
         flash('No file part')
-        return redirect(request.url)
+        return resp
 
     file = request.files['file']
     # if user does not select file, browser also
     # submit an empty part without filename
     if file.filename == '':
         flash('No selected file')
-        return redirect(request.url)
+        return resp
     if file and allowed_file(file.filename):
         filename = secure_filename(file.filename)
         warcPath = os.path.join(app.config['UPLOAD_FOLDER'], filename)
@@ -115,7 +118,9 @@ def upload_file():
         app.cdxjFileContents = getIndexFileContents(app.cdxjFilePath)
 
         # TODO: Release semaphore lock
-        return redirect(request.referrer)
+        resp.location = request.referrer
+
+        return resp
 
 
 @app.route('/ipwbassets/<path:path>')
@@ -163,7 +168,10 @@ def showMementosForURIRs_sansJS():
     urir = request.args.get('url')
     if urir is None or urir.strip() == '':
         return Response('Searching for nothing is not allowed!', status=400)
-    return redirect('/memento/*/' + urir, code=301)
+    resp = redirect('/memento/*/' + urir, code=301)
+    resp.autocorrect_location_header = False
+
+    return resp
 
 
 @app.route('/memento/*/<path:urir>')
@@ -182,7 +190,11 @@ def showMementosForURIRs(urir):
     if len(cdxjLinesWithURIR) == 1:
         fields = cdxjLinesWithURIR[0].split(' ', 2)
         redirectURI = '/memento/{1}/{0}'.format(unsurt(fields[0]), fields[1])
-        return redirect(redirectURI, code=302)
+
+        resp = redirect(redirectURI, code=302)
+        resp.autocorrect_location_header = False
+
+        return resp
 
     msg = ''
     if cdxjLinesWithURIR:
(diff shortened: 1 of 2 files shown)


ipwb/replay.py
```diff
@@ -89,16 +89,19 @@ def allowed_file(filename):
 @app.route('/upload', methods=['POST'])
 def upload_file():
     # check if the post request has the file part
+    resp = redirect(request.url)
+    resp.autocorrect_location_header = False
```
ibnesayeed: I think we can avoid the duplication of logic in different places by simply placing `resp.autocorrect_location_header = False` in an `@app.after_request` wrapped function.

ibnesayeed: This means, reverting all these changes and adding this one line in the `setServerHeader` function.

machawk1: @ibnesayeed Have you tested this? Your correction is inconveniently timely.

machawk1: Also, do you foresee any scenarios where this might have ill effects when applied to responses that are not redirects?

machawk1:
Most of these issues in the replay tests have to do with the test expecting an absolute localhost URL. This can be resolved with the updated relative URI location header.

ipwb/replay.py
ibnesayeed: > Have you tested this? Your correction is inconveniently timely.

No, I have not tested it, but the purpose of `before_request` and `after_request` decorators is to place general logic in one place and keep the code DRY.

> Also, do you foresee any scenarios where this might have ill effects when applied to responses that are not redirects?

No, I don't think it will affect requests that do not have `Location` header in them.

ibnesayeed:
I don't necessarily agree on piling up code duplication and offload it for refactoring later while a simpler approach can easily be tested in the first place, but I will leave that up to you.

machawk1:
@ibnesayeed That decorator is new to me but I think it would be a good improvement after we get the issue fixed. Are there any situations where we may need conformance assurance?

ibnesayeed:
No, that decorator is not new in this repo. We are already using it to add our custom `Server` header in each response. It was just a matter of adding one line in that same function where we include the `Server` header. It would have been `1 addition` instead of `24 additions and 6 deletions`. In my opinion, every line added to the repo is a liability and every duplicate code line is even more evil when it comes to maintenance.

PR #591, opened 2018-10-15
Fix IPFS webui link

### LINKED ISSUE #567: Link to IPFS WebUI is broken
In 00c811a/v0.2018.09.19.1858, it points to http://127.0.0.1:5001/ipwbassets

machawk1:
@ibnesayeed Note that this will not be exposed if not mapped with `-p` when running the docker container. Assuming the ODUCS instance is not using the remapping of port 5001, this should not introduce any sort of security issue. It also allows users to directly access the IPFS webui once again.

Closes #567

### DIFF
--- a/ipwb/assets/webui.js
+++ b/ipwb/assets/webui.js
@@ -188,7 +188,7 @@ function getIPFSWebUIAddress () {
   }
   const fail = function () { console.log('fail') }
   const err = function () { console.log('err') }
-  makeAnAJAXRequest('/ipwbconfig/openEndedPlaceHolder', setIPFSWebUILink, fail, err)
+  makeAnAJAXRequest('/ipfsdaemon/getwebuilink', setIPFSWebUILink, fail, err)
 }
 
 function updateIPFSDaemonButtonUI () {
--- a/ipwb/assets/webui.js
+++ b/ipwb/assets/webui.js
@@ -188,7 +188,7 @@ function getIPFSWebUIAddress () {
   }
   const fail = function () { console.log('fail') }
   const err = function () { console.log('err') }
-  makeAnAJAXRequest('/ipfsdaemon/getwebuilink', setIPFSWebUILink, fail, err)
+  makeAnAJAXRequest('/ipfsdaemon/webuilink', setIPFSWebUILink, fail, err)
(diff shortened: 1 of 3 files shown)


ipwb/replay.py
```diff
@@ -553,7 +553,7 @@ def all_exception_handler(error):
 # the future.
 @app.route('/ipwbconfig/<requestedSetting>')
 def getRequestedSetting(requestedSetting):
-    return Response(ipwbUtils.getIPFSAPIHostAndPort() + '/ipwbassets')
+    return Response(ipwbUtils.getIPFSAPIHostAndPort() + '/webui')
```
ibnesayeed: The comment above this block is misleading and the `requestedSetting` parameter is unused.

machawk1: Yeah, you are right. We used to use (or perhaps just anticipated to use) the route for additional setting retrieval. I will re-check on this and make it a bit more semantic.

machawk1:
@ibnesayeed Can you give the updated changes another review. We were only using that endpoint to fetch the IPFS daemon webui link. I rerouted it to use the existing `ipfsdaemon` endpoint and removed the exclusion from the SW. I tested this locally with a fresh SW.

ipwb/assets/webui.js
```diff
@@ -188,7 +188,7 @@ function getIPFSWebUIAddress () {
   }
   const fail = function () { console.log('fail') }
   const err = function () { console.log('err') }
-  makeAnAJAXRequest('/ipwbconfig/openEndedPlaceHolder', setIPFSWebUILink, fail, err)
+  makeAnAJAXRequest('/ipfsdaemon/getwebuilink', setIPFSWebUILink, fail, err)
```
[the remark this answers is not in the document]
machawk1: I thought about that as well, @ibnesayeed, but was not sure due to the `start` and `stop` "actionables" at that endpoint having to do with controlling the daemon. However, we do have a `status` with an implied GET, so it already has some REST flavor at the endpoint.

machawk1:
@ibnesayeed `webui` endpoint changed to not include `get` in both places. Please let me know if any further changes are needed.

PR #609, opened 2019-05-27
[WIP] Update ipfsapi dependency to new name, adapt ipwb tests to Py3

machawk1:
* Update py-ipfs-api to use its new name.
* Fix client invocation per the new API module's docs.

Fixes #608

PR #611, opened 2019-06-21
Fix semantics of host/port as @ibnesayeed recommended in #609

### LINKED ISSUE #610: Fix semantics of IPFS daemon identifier
In #609 with the integration of the renamed ipfsapi python module, we were required to update using HOST and PORT for parameters to a multiaddress (i.e., beginning with `/`).

In #609 @ibnesayeed noted that HOST implies a hostname where sometimes an IP might be used instead. The distinction of host vs IP for this address is significant, as the multiaddress for each would be different (see [Important changes from ipfs 0.4.x](https://github.com/ipfs/py-ipfs-api#important-changes-from-ipfsapi-04x)), e.g.,

* ipfshttpclient.connect("/dns/<host>/tcp/<port>/http") (for hostnames such as localhost)
* ipfshttpclient.connect("/ip4/<IP-address>/tcp/<port>/http") (for IPv4 addresses)

If an IPFS daemon is remote (#349), this distinction is again important.

Let's modify this to be a single string of host/ip:port and interpret this string on usage, building the multiaddress in the manner that is applicable to the specified host/ip:port.

machawk1:
Fixes #610

### DIFF
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -8,18 +8,18 @@ from .__init__ import __version__ as ipwbVersion
 # ipwb modules
 from . import replay
 from . import indexer
-from . import util
+from . import util as ipwbUtil
 
 from .util import IPWBREPLAY_HOST, IPWBREPLAY_PORT
 
 
 def main():
-    util.checkForUpdate()
+    ipwbUtil.checkForUpdate()
     args = checkArgs(sys.argv)
 
 
 def checkArgs_index(args):
-    if not util.isDaemonAlive():
+    if not ipwbUtil.isDaemonAlive():
         sys.exit()
     encKey = None
     compressionLevel = None
@@ -138,7 +138,7 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-d', '--daemon',
         help='Location of ipfs daemon (default 127.0.0.1:5001)',
-        default=util.IPFSAPI_ADDRESS,
+        default=ipwbUtil.IPFSAPI_ADDRESS,
         dest='daemon_address')
     parser.add_argument(
         '-v', '--version', help='Report the version of ipwb', action='version',
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -137,7 +137,7 @@ def checkArgs(argsIn):
 
     parser.add_argument(
         '-d', '--daemon',
-        help='Location of ipfs daemon (default 127.0.0.1:5001)',
+        help='Multi-address of IPFS daemon (default /dns/localhost/tcp/5001/http)',
         default=ipwbUtil.IPFSAPI_MUTLIADDRESS,
         dest='daemon_address')
     parser.add_argument(
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -138,7 +138,7 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-d', '--daemon',
         help='Location of ipfs daemon (default 127.0.0.1:5001)',
-        default=ipwbUtil.IPFSAPI_ADDRESS,
+        default=ipwbUtil.IPFSAPI_MUTLIADDRESS,
         dest='daemon_address')
     parser.add_argument(
         '-v', '--version', help='Report the version of ipwb', action='version',
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -10,7 +10,7 @@ from . import replay
 from . import indexer
 from . import util
 
-from .util import IPFSAPI_HOST, IPFSAPI_PORT, IPWBREPLAY_HOST, IPWBREPLAY_PORT
+from .util import IPWBREPLAY_HOST, IPWBREPLAY_PORT
 
 
 def main():
@@ -138,7 +138,7 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-d', '--daemon',
         help='Location of ipfs daemon (default 127.0.0.1:5001)',
-        default='{0}:{1}'.format(IPFSAPI_HOST, IPFSAPI_PORT),
+        default=util.IPFSAPI_ADDRESS,
         dest='daemon_address')
     parser.add_argument(
         '-v', '--version', help='Report the version of ipwb', action='version',
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -137,7 +137,8 @@ def checkArgs(argsIn):
 
     parser.add_argument(
         '-d', '--daemon',
-        help='Multi-address of IPFS daemon (default /dns/localhost/tcp/5001/http)',
+        help=("Multi-address of IPFS daemon "
+              "(default /dns/localhost/tcp/5001/http)"),
         default=ipwbUtil.IPFSAPI_MUTLIADDRESS,
         dest='daemon_address')
     parser.add_argument(
--- a/ipwb/indexer.py
+++ b/ipwb/indexer.py
@@ -52,7 +52,11 @@ from .__init__ import __version__ as ipwbVersion
 
 DEBUG = False
 
-IPFS_API = ipfsapi.Client(ipwbUtils.IPFSAPI_MUTLIADDRESS)
+try:
+    IPFS_API = ipfsapi.Client(ipwbUtils.IPFSAPI_MUTLIADDRESS)
+except ipfsapi.exceptions.AddressError:
+    print('Malformed multiaddress for the daemon')
+    sys.exit()
 
 
 def s2b(s):  # Convert str to bytes, cross-py
--- a/ipwb/indexer.py
+++ b/ipwb/indexer.py
@@ -52,10 +52,9 @@ from .__init__ import __version__ as ipwbVersion
 
 DEBUG = False
 
-try:
-    IPFS_API = ipfsapi.Client(ipwbUtils.IPFSAPI_MUTLIADDRESS)
-except ipfsapi.exceptions.AddressError:
-    print('Malformed multiaddress for the daemon')
+IPFS_API = ipwbUtils.createIPFSClient()
+if IPFS_API is None:
+    print("Error initializing IPFS API client")
     sys.exit()
 
 
--- a/ipwb/indexer.py
+++ b/ipwb/indexer.py
@@ -33,8 +33,8 @@ from six.moves import input
 from six import PY2
 from six import PY3
 
-from .util import IPFSAPI_HOST, IPFSAPI_PORT
 from .util import iso8601ToDigits14
+from . import util as ipwbUtils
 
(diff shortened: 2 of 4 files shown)


ipwb/__main__.py
```diff
@@ -138,7 +138,7 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-d', '--daemon',
         help='Location of ipfs daemon (default 127.0.0.1:5001)',
-        default='{0}:{1}'.format(IPFSAPI_HOST, IPFSAPI_PORT),
+        default=util.IPFSAPI_ADDRESS,
```
ibnesayeed: I don't think we should rely on old host and port style. We can change this argument to accept multi-address which will simplify many things internally and will be more flexible to connect to IPFS nodes in many different ways.

Also, `util` is imported as `ipwbUtils` so we can make this one to be same as the rest.

ipwb/util.py
```diff
@@ -39,9 +42,26 @@
...
+        (host, port) = hostAndPort.split(':')
+    except ValueError as err:
+        # Naive check to determine if multiaddr passed in
+        if hostAndPort[0] == '/' and hostAndPort.count('/'):
+            return hostAndPort
+        else:
+            print('Error reading IPFSAPI address')
+            sys.exit()
+    try:
+        socket.inet_aton(host)
```
ibnesayeed: If we attempt to check the connection using the `ipfshttpclient` itself, then we will not need `socket` here and more importantly we will have no need to sore the `IPFSAPI_ADDRESS`. This whole `getMultiAddress` function can go away and be replaced with a simple constant holding the multi-address.

machawk1: The socket is usage is just a means for IP address validation. If it's not, the exception will be thrown.

The multiaddress will vary based on whether it's an IP address or host, per the `getMultiAddress` logic.

ipwb/__main__.py
machawk1: Made the `ipwbUtils` variable naming consistent in 8c28301.

machawk1: Specifying a multiaddress on the command line seems really awkward for a user, particularly with the variable rules based on whether it's an IP or a host. I'd rather the mutliaddress be inferred based on the IP and host that the user passes in.

ipwb/util.py
ibnesayeed: > The socket is usage is just a means for IP address validation. If it's not, the exception will be thrown.

I understand the purpose and that is what I meant by `If we attempt to check the connection using the ipfshttpclient itself...` which can ensure that the can indeed connect to the provided multi-addr. If there is anything wrong, the client library will throw and exception. Using the client may come with many other corner cases covered. This custom validation is both limiting and incomplete. It enforces us to only allow a host and port or IPv4 and port style connection and nothing else, but we should leverage the flexibility we get from the multi-addr and offload any validation to the library itself.

machawk1: @ibnesayeed Instead of the logic in `getMultiAddress()` do you think we should instantiate an ipfshttpclient `client` instance, like we usually do, and if it can't connect (as indicated by a timeout, exception, error, etc), try a different multiaddress pattern?

The point of the function is to determine if the "host" is a hostname or IP address and generate the appropriately formatted mutliaddress.

If a client specifies an IP address or hostname on the CLI, the resulting multiaddress formatting will vary based on the "host".

ibnesayeed: In the very first comment in the `__main__.py` file I suggested that we simply do not accept daemon address as IP addr/hostname + port number, Instead, we ask the user to supply a complete and valid multi-addr. This way, we can simply store the default multi-addr as a plain string and allow it to be overwritten completely by the CLI argument (if present). We have no business of storing it in pieces or validating in pieces. In order to validate the multi-addr, simply attempt to connect to it by creating a client using the function provided by the IPFS library. If there is any formatting or network issue, the library will through an exception that can be caught and reported. In case of success, we may choose to reuse this client or dispose it. This method does not require us to attempt for many different ways a multi-addr can be formatted, because we won't constructing it piece by piece.

machawk1: Ok, stand by. I'll implement that logic then give you a ping.

machawk1: @ibnesayeed Please review c9f48d3.

ipwb/util.py
```diff
@@ -42,33 +44,16 @@
...
-        return f"/ip4/{host}/tcp/{port}/http"
-    except socket.error:
-        return f"/dns/{host}/tcp/{port}/http"
-
-
-def isDaemonAlive(hostAndPort=IPFSAPI_ADDRESS):
+def isDaemonAlive(daemonMultiaddr=IPFSAPI_MUTLIADDRESS):
     """Ensure that the IPFS daemon is running via HTTP before proceeding"""
-    client = ipfsapi.Client(getMultiAddress(hostAndPort))
+    client = ipfsapi.Client(daemonMultiaddr)
```
ibnesayeed: I would wrap this client creation line in a try-except block, because this is where any multi-addr formatting issues would be validated. This means, if the user supplies an invalid multi-addr, an appropriate exception will be thrown here.

ipwb/__main__.py
```diff
@@ -138,7 +138,7 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-d', '--daemon',
         help='Location of ipfs daemon (default 127.0.0.1:5001)',
```
ibnesayeed:
```suggestion
        help='Multi-address of IPFS daemon (default /dns/localhost/tcp/5001/http)',
```

ibnesayeed:
Also, we may want to add a call to `isDaemonAlive()` function during argument parsing so that we ensure that any malformed multi-addr argument is caught before the application boots up.

ipwb/indexer.py
```diff
@@ -52,7 +52,11 @@
 
 DEBUG = False
 
-IPFS_API = ipfsapi.Client(ipwbUtils.IPFSAPI_MUTLIADDRESS)
```
ibnesayeed: #DRY!!! Define a `createIPFSClient()` function in the `util.py` file that validates the address. If validation fails (i.e., an exception in thrown), return `None`, otherwise return the client object. In `indexer.py` and `replay.py` files call that function and assign the returned value to a variable. Then check if the variable is `None`, immediately exit. Reuse the same function inside the `isDaemonAlive()` function, but do not exit there, simply return `False` or whatever makes sense in that context.

ipwb/replay.py
```diff
@@ -68,7 +68,11 @@
 app.config['UPLOAD_FOLDER'] = UPLOAD_FOLDER
 app.debug = False
 
-IPFS_API = ipfsapi.Client(ipwbUtils.IPFSAPI_MUTLIADDRESS)
```
ibnesayeed: Read the comment above!

ipwb/util.py
```diff
@@ -46,7 +46,11 @@
 
 def isDaemonAlive(daemonMultiaddr=IPFSAPI_MUTLIADDRESS):
     """Ensure that the IPFS daemon is running via HTTP before proceeding"""
-    client = ipfsapi.Client(daemonMultiaddr)
```
ibnesayeed: Read the comment above!

ipwb/util.py
```diff
@@ -6,6 +6,7 @@
 import os
 import sys
 import requests
+import socket
```
ibnesayeed:
```suggestion
```

This import is not needed anymore!

ipwb/util.py
```diff
@@ -18,17 +19,22 @@
 from six.moves.urllib.request import urlopen
 import json
 from .__init__ import __version__ as ipwbVersion
+from . import util as ipwbUtils
```
ibnesayeed:
```suggestion
```

No need to import a module from within!

ipwb/util.py
```diff
@@ -39,16 +45,26 @@
...
+    try:
+        return ipfsapi.Client(IPFSAPI_MUTLIADDRESS)
+    except ipfsapi.exceptions.AddressError:
+        return None  # Malformed multiaddress for the daemon
+
+
+def isDaemonAlive(daemonMultiaddr=IPFSAPI_MUTLIADDRESS):
     """Ensure that the IPFS daemon is running via HTTP before proceeding"""
-    client = ipfsapi.Client(f"/dns/{IPFSAPI_HOST}/tcp/{IPFSAPI_PORT}/http")
+    client = ipwbUtils.createIPFSClient()
```
ibnesayeed:
```suggestion
    client = createIPFSClient()
```

It is a local function.

machawk1: Yeah, noticed that after I pushed. Fixed in 05e7eba.

ipwb/util.py
machawk1: See below.

ipwb/util.py
machawk1: Fixed in 911ae6f.

ipwb/util.py
```diff
@@ -39,16 +43,26 @@
 dtPattern = re.compile(r"^(\d{4})(\d{2})?(\d{2})?(\d{2})?(\d{2})?(\d{2})?$")
 
 
-def isDaemonAlive(hostAndPort="{0}:{1}".format(IPFSAPI_HOST, IPFSAPI_PORT)):
+def createIPFSClient():
```
ibnesayeed:
```suggestion
def createIPFSClient(daemonMultiaddr=IPFSAPI_MUTLIADDRESS):
```

Following the pattern used in `isDaemonAlive()`, it perhaps is better to also make it parametrized with the default value in place.

ipwb/util.py
```diff
@@ -39,16 +43,26 @@
...
+    try:
+        return ipfsapi.Client(IPFSAPI_MUTLIADDRESS)
```
ibnesayeed:
```suggestion
        return ipfsapi.Client(daemonMultiaddr)
```

Per above!

ipwb/util.py
```diff
@@ -39,16 +43,26 @@
...
+    except ipfsapi.exceptions.AddressError:
```
ibnesayeed: Is this the right exception here? I have seen many different exception being raised based on the argument.

machawk1: That's the one I encountered when entering bad multiaddresses.

ibnesayeed: We might make it more generic here unless we are confident about the exception chain.

ipwb/util.py
machawk1: Addressed in 1d57554.

ipwb/util.py
machawk1: 1d57554

ipwb/util.py
machawk1: Added a second in bdecf23. We don't want to reinvent the ipfsapi test module, so the two ought to be sufficient for now.

PR #615, opened 2019-07-25
Bump flask version while we figure out why we can't yet use latest

machawk1:
Ping @ibnesayeed 

Flask 1.0 breaks things but this update still passes tests. Still trying to figure out what will get us to latest.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 warcio>=1.5.3
 ipfshttpclient>=0.4.12
-Flask==0.12.3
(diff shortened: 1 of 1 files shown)


ibnesayeed:
I am not sure if this version mitigates the vulnerability though.

machawk1:
@ibnesayeed I do not think it does, but it gets us closer to latest. I created a new ticket in #616 to hopefully get us updated with Flask and resolve the vulnerability. Also relevant: #617. Feel free to chime in on either.

PR #618, opened 2019-07-25
Update to latest Flask

### LINKED ISSUE #616: Update to Flask ≥ 1.0
ipwb currently uses Flask version 0.12.4. #613 was an attempt by a bot to update the repo with a newer version to solve a (still present) security issue. This caused ipwb tests to fail due to some difference between 0.12.3 (at the time) and 1.0.

This ticket represents an effort to move to a newer version of Flask.

machawk1:
Closes #616

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -271,8 +271,7 @@ def showMemento(urir, datetime):
     else:
         resp = show_uri(uri, newDatetime)
 
-    resp.headers.add('Link', linkHeader)
-    # resp.headers['Link'] = linkHeader
+    resp.headers['Link'] = linkHeader
 
     return resp
 
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -271,7 +271,8 @@ def showMemento(urir, datetime):
     else:
         resp = show_uri(uri, newDatetime)
 
-    resp.headers['Link'] = linkHeader
+    resp.headers.add('Link', linkHeader)
+    # resp.headers['Link'] = linkHeader
 
     return resp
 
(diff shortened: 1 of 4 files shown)


ipwb/replay.py
```diff
@@ -271,7 +271,8 @@ def showMemento(urir, datetime):
     else:
         resp = show_uri(uri, newDatetime)
 
-    resp.headers['Link'] = linkHeader
+    resp.headers.add('Link', linkHeader)
```
ibnesayeed: This is inconsistent with how other instances of `resp.headers` are updated in the replay. Perhaps you should investigate why it works in other places without and explicit `.add()` method, but not here.

machawk1: I agree, there should be some normalization to allow for the predictable manipulation of the type used as a "Response" here. Some more tests are needed to replicate the other instances of Response fabrication.

machawk1:
Per the sub-comment above:

```
    if newDatetime != datetime:
        resp = redirect('/memento/{0}/{1}'.format(newDatetime, urir), code=302)
        print(type(resp))
    else:
        resp = show_uri(uri, newDatetime)
        print(type(resp))
```
Initial conditional produces a `werkzeug.wrappers.response.Response`. Else clause produces a `flask.wrappers.Response`. This might be the reason that the result from the else is a tuple and not an object, thus the need for `resp.headers.add('Link', linkHeader)` instead of `resp.headers['Link'] = linkHeader`, which resulted in an exception for the tests that executed the else clause.

machawk1:
I reverted back to the latter snippet in the previous message and the else clause no longer throws an exception. Maybe this was a bug in Flask 1.0 that is not present in 1.1.

PR #631, opened 2020-03-01
Implement Binary Search in Replay

### LINKED ISSUE #604: Implement memory-efficient in-file binary search for CDXJ indexes
Currently, the complete index is loaded in memory for faster lookup, but not scalable for large index files as the memory usage grows linearly with the size of index. We can instead implement an in-file binary search to make the memory usage constant for any arbitrary index file size while still being efficient in lookup due to the nature of binary search. A similar implementation is present in the newly introduced [MementoMap](https://github.com/oduwsdl/MementoMap) repo.

machawk1:
This implementation is heavily inspired by @ibnesayeed's [MementoMap implementation](https://github.com/oduwsdl/mementomap).

In the future we _might_ align this with @ibnesayeed [upcoming API](https://github.com/ibnesayeed/binsearch) but for now, this drastically speeds up replay and thus closes #604. I recommend we merge then refine the implementation.

@ibnesayeed Can you have a look and let me know of some refinements with regard to semantics and implementation?


ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+    mobj.close()
+
+
+def lookup(iter, surt, datetime=None):
+    for idx, key in enumerate(lookup_keys(surt)):
+        res = bin_search(iter, key.encode(), datetime)
+        if res:
+            return res
+
+
```
ibnesayeed: What are these functions doing here? These were specific to MementoMap where I had to find partial key matches.

```suggestion
```

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+        try:
+            m = keyre.match(key)
+            keys.append(f"{m[1]}{m[2]}*")
+            key = m[1]
+        except Exception as e:
+            key = key.strip("/,")
+
+    return keys
+
+
```
ibnesayeed: This does not belong here either? This was specific to MementoMap where I had to create all partial key for lookup.

```suggestion
```

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+def getCDXJLinesWithURIR(urir, indexPath, datetime=None):
+    """ Get all CDXJ records corresponding to a URI-R """
+    if not indexPath:
+        indexPath = ipwbUtils.getIPWBReplayIndexPath()
+    indexPath = getIndexFileFullPath(indexPath)
+
+    # Convert URI-R to surt
+    surtedURIR = surt.surt(urir, path_strip_trailing_slash_unless_empty=True)
+
+    res = run_batchlookup(surtedURIR, indexPath, datetime)
```
ibnesayeed: After removing the `run_batchlookup` function, we need to think about it and reimplement in a way that is efficient and works.

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+            right = mid
+            continue
+
+        surtk, rest = line.split(maxsplit=1)
+        datetimeK = rest.split()[0].decode()
+        if surtk[-1:] == b'/':
+            surtk = surtk[0:-1]
+
+        # TODO: find a more elegant way for comparison than manually
+        # trimming the last two chars off of the surtk
```
ibnesayeed: This will strip any number of consecutive trailing slashes:

```suggestion
        surtk = surtk.rstrip(b"/")
```

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+def bin_search(iter, key, datetime=None):
+    # Read line encompassing current position
+    ln = iter.readline()
+
+    while ipwbUtils.isCDXJMetadataRecord(ln):
+        ln = iter.readline()
+
+    surtk, rest = ln.split(maxsplit=1)
+    if key == surtk:
+        return [ln]
```
ibnesayeed: If the first non-metadata line matches just the key, will you consider it a match?

* What about the datetime?
* What about the potential trailing slash in the surtk?

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+        iter.seek(mid)
+        iter.readline()  # Purge rest of current line
+        line = iter.readline()  # Read the next full line
+
+        if len(line) == 0:
+            right = mid
+            continue
+
+        surtk, rest = line.split(maxsplit=1)
+        datetimeK = rest.split()[0].decode()
```
ibnesayeed:
```suggestion
        surtk, datetimeK, rest = line.split(maxsplit=2)
        datetimeK = datetimeK.decode()
```

Ideally, both the supplied key and datetime should be bytesarray, so that we do not need to decode each line read from the index file.

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+        if surtk[-1:] == b'/':
+            surtk = surtk[0:-1]
+
+        # TODO: find a more elegant way for comparison than manually
+        # trimming the last two chars off of the surtk
+
+        if key == surtk:
+            if datetime and datetime != datetimeK:
```
ibnesayeed: I think you wanted to check for equality, not otherwise?

```suggestion
            if datetime and datetime == datetimeK:
```

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+
+                break
+            lines.add(line)
+            # Iterate further to get lines after selection point
+            nextLine = iter.readline()
+            while nextLine:
+                surtk, rest = nextLine.split(maxsplit=1)
+                datetimeK = rest.split()[0].decode()
+                if surtk[-1:] == b'/':
+                    surtk = surtk[0:-1]
```
ibnesayeed:
```suggestion
                surtk, datetimeK, rest = nextLine.split(maxsplit=2)
                datetimeK = datetimeK.decode()
                surtk = surtk.rstrip(b"/")
```

ipwb/replay.py
```diff
@@ -179,6 +179,126 @@ def showMementosForURIRs_sansJS():
...
+
+            # Continue searching until find first instance
+            right = mid
+        elif key > surtk:
+            left = mid
+        else:
+            right = mid
+
+    # Convert uniq set to list then sort and return
+    ret = sorted(list(lines))
```
ibnesayeed: I am not really happy about holding all the matches in memory and sort them if we can find the first match and iterate from there more efficiently, especially, if a single URI is captured way too many times. However, if it works for now, it's alright.

ipwb/util.py
```diff
@@ -145,6 +145,10 @@ def setLocale():
 
 def digits14ToRFC1123(digits14):
     setLocale()
+    # Trim to 14-digits max for testing
+    print(digits14)
```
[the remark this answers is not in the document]
machawk1: Yes, it should be removed.

ipwb/replay.py
machawk1: Coded in similar paths through the code. Will simplify calls and remove.

ipwb/replay.py
machawk1: Yes, you are correct. Odd that this is the inverse logic but still worked. I will look into it further.

ipwb/replay.py
machawk1: Your way is better, will apply.

ipwb/replay.py
ibnesayeed: In MementoMap, I did it this way because the need there was to find the longest common sub-URI-key. So, for a given surt `foo,bar)/baz/blah`, I was supposed to check for any matches in the following order and terminate as soon as there is a match:

```
foo,bar)/baz/blah
foo,bar)/baz/blah/*
foo,bar)/baz/*
foo,bar)/*
foo,bar,*
foo,*
```

In this case, `lookup_keys()` was returning such a list of partial keys that was then binary searched.

ipwb/replay.py
machawk1: @ibnesayeed In the future, should we:
* find the first match
* count the number of match
* return the `fh` iterator pointing to the first match and count

This would prevent retaining the matched lines. I am open to other approaches.

ipwb/replay.py
machawk1: Turns out it was not working and giving us false positives of URI-Ms resulting in the wrong memento being returned. Fixed in a603cf5.

ipwb/replay.py
machawk1: Fixed in a603cf5.

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
+
+    return ret
+
+
+def getMatchDegree(surt, datetime, surtK, datetimeK):
+    if surt == surtK:
```
ibnesayeed:
```suggestion
    if surt == surtK:
        datetimeK = datetimeK.decode()
```

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
 
+def bin_search(iter, key, datetime=None):
+    # Read line encompassing current position
+    ln = iter.readline()
+
+    while ipwbUtils.isCDXJMetadataRecord(ln):
+        ln = iter.readline()
+
+    surtk, datetimeK, rest = ln.split(maxsplit=2)
+    datetimeK = datetimeK.decode()
```
ibnesayeed: DRY
```suggestion
```

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
+        iter.seek(mid)
+        iter.readline()  # Purge rest of current line
+        line = iter.readline()  # Read the next full line
+
+        if len(line) == 0:
+            right = mid
+            continue
+
+        surtk, datetimeK, rest = line.split(maxsplit=2)
+        datetimeK = datetimeK.decode()
```
ibnesayeed: DRY
```suggestion
```

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
+        matchDegree = getMatchDegree(key, datetime, surtk, datetimeK)
+
+        if matchDegree == MementoMatch.RIGHTKEYWRONGDATE:
+            lines.add(line)
+            # Iterate further to get lines after selection point
+            nextLine = iter.readline()
+            while nextLine:
+                surtk, datetimeK, rest = nextLine.split(maxsplit=2)
+                surtk = surtk.rstrip(b"/")
+                datetimeK = datetimeK.decode()
```
ibnesayeed: DRY
```suggestion
```

ipwb/replay.py
machawk1: Good catch: the decoding can be done in the called function. Will apply.

ipwb/replay.py
machawk1: Applied in af4aa1f

ipwb/replay.py
machawk1: Applied in af4aa1f.

ipwb/replay.py
machawk1: Applied in af4aa1f.

ipwb/replay.py
machawk1: Applied in af4aa1f.

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
+
+                matchDegree = getMatchDegree(key, datetime, surtk, datetimeK)
+                if matchDegree == MementoMatch.RIGHTKEYWRONGDATE:
+                    lines.add(nextLine)
+                elif matchDegree == MementoMatch.EXACTMATCH:
+                    # Exact match found while iterating
+                    lines.clear()
+                    lines.add(nextLine)
+                    break
```
ibnesayeed: I think the logic here needs to be reworked and changed from `break` to `return` as the outer loop will continue to execute, as a result, despite clearing any prior non-exact matches, more non-exact matches might be added to the set that follow the exact match.

```suggestion
                elif matchDegree == MementoMatch.EXACTMATCH:
                    # Exact match found while iterating
                    return [nextLine]
```

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
+
+                nextLine = iter.readline()
+
+            # Continue searching until find first instance
+            right = mid
+        elif matchDegree == MementoMatch.EXACTMATCH:
+            lines.clear()
+            lines.add(line)
+
+            break
```
ibnesayeed: While this one will break the outer loop, you might to use return here as well for similarity sake.

```suggestion
        elif matchDegree == MementoMatch.EXACTMATCH:
            return [line]
```

ipwb/replay.py
```diff
@@ -179,6 +180,111 @@ def showMementosForURIRs_sansJS():
...
+                matchDegree = getMatchDegree(key, datetime, surtk, datetimeK)
+                if matchDegree == MementoMatch.RIGHTKEYWRONGDATE:
+                    lines.add(nextLine)
+                elif matchDegree == MementoMatch.EXACTMATCH:
+                    # Exact match found while iterating
+                    lines.clear()
+                    lines.add(nextLine)
+                    break
+
+                nextLine = iter.readline()
```
ibnesayeed: This is a problematic thing (you might want to test it against a large CDX), because I feel that if the SURT key matched once without any exact match, this iterator is not going to terminate the inner loop and will process the whole index from that point till the end. The `WRONGKEY` condition needs to be checked and inner loop needs to be broken based on that to avoid it.

ipwb/replay.py
machawk1: I agree. The programmatic flow changed here with regard to manipulating the progressively built `set()`. Change applied in 854af87.

ipwb/replay.py
machawk1: Applied in 3bc6e37.

ipwb/replay.py
machawk1: @ibnesayeed Please verify logic here some logic I am proposing:

If we break from the while loop with an `elif matchDegree == MementoMatch.WRONGKEY`, it should hold true that all matched surted URIs have being exhausted, as the conditional preceding the while loop already confirmed the `matchDegree == MementoMatch.RIGHTKEYWRONGDATE` condition.

I am trying to prevent the iteration from prematurely terminating, so would like the sanity check. :)

ibnesayeed: Yes, that's what I suggested.

machawk1: Implemented in e3a85eb.

machawk1:
I fabricated (attached) a 2500 entry CDXJ with all entries being the same URI-R (`memento.us`) and started ipwb replay from e3a85eb. Everything loads sufficiently quick despite all entries being in memento (per https://github.com/oduwsdl/ipwb/pull/631#discussion_r386159465). A larger sample is needed to verify this.

One caveat, when I dereferenced the URI-M, the web UI returned `Error` and the CLI showed an exception in the console that I chased down to `payload = IPFS_API.cat(digests[-1])`. This will need to be re-checked to have a less erroneous UI/UX if the entry is not found in IPFS.

[sample_sorted.cdxj.txt](https://github.com/oduwsdl/ipwb/files/4281973/sample_sorted.cdxj.txt)

ipwb/replay.py
```diff
@@ -179,6 +180,107 @@ def showMementosForURIRs_sansJS():
...
+        ln = iter.readline()
+
+    surtk, datetimeK, rest = ln.split(maxsplit=2)
+
+    matchDegree = getMatchDegree(surt, datetime, surtk, datetimeK)
+    if matchDegree == MementoMatch.EXACTMATCH:
+        return [ln]
+
+    # If further searching required...
+    left = 0
```
ibnesayeed: This initial section has a few issues:

* I the `RIGHTKEYWRONGDATE` match occurs at the first data line then it will not be included in the response as the while loop below will already advance to the second data line.
* However, it may not be happening because the last line in the select sets the `left` to `0` (i.e., brings the start of the search to the beginning), which means all this metadata skipping effort is rendered useless.
* We really do not need to worry about skipping metadata lines as such because for a usual SURT key those lines will not match any way.
* We should also add an exception handling for `ValueError: not enough values to unpack (expected 3, got 2)` when performing split and assigning it to a fixed number of variables, because some metadata lines might not have two spaces in them, if we chose to perform bin search on the whole file without filtering any metadata lines out.

That said, following changes should allow skipping metadata lines and setting the start boundary properly:

```suggestion
    # Skip metadata lines
    while iter.peek(1)[:1] == b'!':
        iter.readline()

    # Set the beginning position to the start of the first data line
    left = iter.tell()
```

ipwb/replay.py
```diff
@@ -179,6 +180,107 @@ def showMementosForURIRs_sansJS():
...
+        mid = (right + left) // 2
+        iter.seek(mid)
+        iter.readline()  # Purge rest of current line
+        line = iter.readline()  # Read the next full line
+
+        if len(line) == 0:
+            right = mid
+            continue
+
+        surtk, datetimeK, rest = line.split(maxsplit=2)
```
ibnesayeed: To handle malformed lines safely:

```suggestion
        try:
            surtk, datetimeK, rest = line.split(maxsplit=2)
        except ValueError as e:
            continue
```

ipwb/replay.py
```diff
@@ -179,6 +180,107 @@ def showMementosForURIRs_sansJS():
...
+
+        surtk = surtk.rstrip(b"/")
+        key = key.rstrip(b"/")
```
ibnesayeed: Move this `key` sanitization outside of the while loop so it is done once as it is a received function argument, not something that is generated on each line like `surtk`.

ipwb/replay.py
```diff
@@ -179,6 +180,107 @@ def showMementosForURIRs_sansJS():
...
+
+        matchDegree = getMatchDegree(key, datetime, surtk, datetimeK)
+
+        if matchDegree == MementoMatch.RIGHTKEYWRONGDATE:
+            lines.add(line)
+            # Iterate further to get lines after selection point
+            nextLine = iter.readline()
+            while nextLine:
+                surtk, datetimeK, rest = nextLine.split(maxsplit=2)
```
ibnesayeed: This is less likely to hit a malformed line, but you can protect this too in a try block like the suggestion above.

ipwb/replay.py
machawk1: @ibnesayeed Outside of the preceding while loop?

https://github.com/oduwsdl/ipwb/blob/8cb614408d740670a923abc992169fb076630df5/ipwb/replay.py#L203-L216

This snippet looks like that cannot be done due to the dependency on the parameters of the while loop established just before finishing each iteration and reuse on the next.

ipwb/replay.py
machawk1: Applied these in https://github.com/oduwsdl/ipwb/tree/issue-604-metadata for tests to be run.

ipwb/replay.py
ibnesayeed: It is not clear to me why we cannot move `key = key.rstrip(b"/")` line before the outer loop starts. In fact we should be able to place it immediately after function body begins.

machawk1: Ah, ok, you are correct, @ibnesayeed . `key` sanitization can be moved before the outer loop. I misread that originally as both `key` and `surtk` sanitization, the latter of which depends on operations within the `while`.

ipwb/replay.py
machawk1: A couple of the tests in TravisCI failed with this change, reporting the unexpected error code. This will be a good opportunity to verify the expected status code and that the revised binsearch algorithm is returning the expected representation.

ipwb/replay.py
ibnesayeed: I thought I made the difference in `key` and `surtk` clear in the first comment. :)

machawk1: Applied in fa1b68c.

machawk1: @ibnesayeed You did. I misinterpreted `key` for key where the latter might refer to both `surtk` and `key`.

ipwb/replay.py
ibnesayeed: I think we should focus on unit tests for this binsearch implementation to catch issues around corner cases, instead of integration test.

machawk1: I agree, but these integration tests should pass, too. Perhaps some unit tests would help remedy the issues. It might help with coverage, too.

machawk1: @ibnesayeed In planning these unit tests, should the data be within the test code or read from a file? I foresee enumerating all of the permutations for testing `bin_search` could make for some hard-to-maintain test code at the same time, it would best test the function itself to have the data inline.

ibnesayeed: > I agree, but these integration tests should pass, too. Perhaps some unit tests would help remedy the issues. It might help with coverage, too.

Integration tests are useful and they should pass. They catch many issues, but they are generally not very effective in isolating the root causes.

machawk1: `bin_search` expects a file handle with its usage of `readline()`, so mocking this with converting a list of strings converted to an iterator might not work...unless there is a way to bind the iterator's `readline()` function (which does not pre-exist) to `next()`.

ibnesayeed: > In planning these unit tests, should the data be within the test code or read from a file? I foresee enumerating all of the permutations for testing bin_search could make for some hard-to-maintain test code at the same time, it would best test the function itself to have the data inline.

I would create an index file that is only geared towards binary search test, not for replay. Then I will place data suitable for various scenarios in it and device test cases against this. I will place unique information in the JSON block to identify the line(s) that should show up (or not show up) in the result of various test cases. Following is a sample that I have put together very quickly, but more lines can be added to it without altering the essence of the existing lines. It is likely that a single file might not be sufficient to test every scenario (for example, if we were to test how will the bin search behave in an empty file or one with a single line), but it will be a good start.

```
!context ["http://oduwsdl.github.io/contexts/memento"]
!meta {"created_at": "2020-03-03T19:13:37", "purpose": "BinarySearch Testing"}
com,cnn)/ 20080117022613 {"desc": "first line"}
com,cnn)/ 20100325193658 {"desc": "second line"}
com,cnn)/ 20100327172049 {"desc": "third line"}
com,cnn)/ 20121114231445 {"desc": "long line", "lorem": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi ullamcorper diam dolor, sed molestie nisl iaculis vitae. Ut faucibus magna dapibus tortor scelerisque ullamcorper. Aenean varius fermentum lorem id placerat. Praesent sed luctus eros. Vestibulum facilisis tristique fermentum. Integer et tempus nunc. Nunc luctus dolor id risus dignissim, a maximus risus facilisis. Nam nec purus malesuada, viverra neque vel, luctus lacus. Phasellus facilisis tempus congue. Maecenas posuere tincidunt nisl, eu vestibulum nisl lobortis ut. Vestibulum porta elit sodales, laoreet dolor quis, volutpat quam. Cras tincidunt imperdiet lacus sit amet rutrum. Sed posuere, ipsum et pretium tempus, orci ex pellentesque ante, eu aliquet augue velit bibendum justo."}
com,cnn)/ 20181221190104 {"desc": "last homepage line"}
com,cnn)/news 20170111124504 {"desc": "single capture line"}
com,cnn)/news?id=ipwb-is-awesome&type=latest+news 20000306162047 {"desc": "line with query param"}
uk,co,bbc)/ 20051222010821 {"desc": "root path"}
uk,co,bbc)/ 20120320091103 {"desc": "first duplicate"}
uk,co,bbc)/ 20120320091103 {"desc": "second duplicate"}
uk,co,bbc)/ 20190121071213 {"desc": "closest past to 20190123071550"}
uk,co,bbc)/ 20190123071559 {"desc": "closest to 20190123071550"}
uk,co,bbc)/ 20200111053342 {"desc": "last capture"}
xyz,example,foo)/bar/baz?blah 20071114090832 {"desc": "last line"}
```

ibnesayeed: > `bin_search` expects a file handle with its usage of `readline()`, so mocking this with converting a list of strings converted to an iterator might not work...unless there is a way to bind the iterator's `readline()` function (which does not pre-exist) to `next()`.

This function is about performing searches in a file, not in the memory, so I would not even think about putting data inline, but a separate index file is what should be used as suggested above. That said, if there is a situation where you do need to mock inline data as a file object, you can use [io.BytesIO](https://docs.python.org/3/library/io.html#io.BytesIO).

machawk1:
> > @ibnesayeed It would be interesting to use your code-in-github-comments GitHub action to evaluate the binsearch speedup in this PR.
> 
> Sure, please go ahead and craft a sequence of commands or python code that you would write to compare the two variations if you were to test it locally on your machine. From there it should be easy to convert it to something that we can make it run here and report the results back.

This could probably be performed using GitHub Actions, right?

ibnesayeed:
@machawk1 any plans to publish `cdxjGenerator` as a package and with a CLI command like `cdxjen`? Currently, it would require cloning the repo, changing the working directory, installing requirements, and then executing the script, but when released as a package and CLI, it will become a 2-step process, `pip install cdxjen && cdxjen`.

machawk1:
@ibnesayeed I thought I had already done so but im writing the above procedure, realized I had not, so created https://github.com/machawk1/cdxjGenerator/issues/3.

machawk1:
Merging this in after multiple discussions and a long delay. It is not the optimal implementation but it is a step in the right direction.

Issue #595, opened 2018-11-01
WARC with query parameter not found in CDXJ


machawk1:
I am using a WARC (attached) and unable to access the mementos within when using ipwb 0.2018.10.15.1346 from source via Docker. Performing the typical indexing procedure shows two captures (pictured). Clicking on a link results in a 404 for the URI-R `biorisk.pensoft.net/articles.php?id=1904` at `20181101192316`.

Strangely, the 404 reports:
`No capture found for biorisk.pensoft.net/articles.php?id=1904?id=1904`
^...note the extra query parameter.

Additionally, the CDXJ and Link TimeMap links on the 404 point to:
`http://localhost:5002/memento/20181101192316/biorisk.pensoft.net/timemap/link/ and `http://localhost:5002/memento/20181101192316/biorisk.pensoft.net/timemap/cdxj/`, which seems to indicate that the URI-R is being parsed incorrectly.

This is a compound issue.

Generated CDXJ:
```
!context ["http://tools.ietf.org/html/rfc7089"]
!meta {"created_at": "2018-11-01T19:30:29.255640", "generator": "InterPlanetary Wayback v.0.2018.10.15.1346"}
net,pensoft,biorisk)/article_preview.php?id=1904 20181101192316 {"locator": "urn:ipfs/QmcF6xX5XUrj19R9eZDTPLc27b4BNz8JH6mu76FUXu5AvP/QmacNjx3oyoy3q9gwwQCj6dv8AH3ybSMe8DVTmKZXmQoK5", "original_uri": "http://biorisk.pensoft.net/article_preview.php?id=1904", "title": "China in the anthropocene: Culprit, victim or last best hope for a global ecological civilisation?", "mime_type": "text/html;charset=utf-8", "status_code": "200"}
net,pensoft,biorisk)/articles.php?id=1904 20181101192316 {"locator": "urn:ipfs/QmVDfudfARdRxKu1pWmhsWrTLM4F4Lfwdw6qJx9K5aaTuA/QmdPDpQD2bLbQvHeJtEUxvyBvav4Tp1GMGMkyMtMU8Ye7w", "original_uri": "http://biorisk.pensoft.net/articles.php?id=1904", "title": "China in the anthropocene: Culprit, victim or last best hope for a global ecological civilisation?", "mime_type": "text/html;charset=utf-8", "status_code": "200"}
[...]

PR #633, opened 2020-04-18
Issue 595: GET arguments support

anatoly-scherbakov:
With GET arguments widely used to address various CSS and JS files (for caching purposes, etc), #595 is quite annoying. The site I was trying to archive is a Russian government website https://стопкоронавирус.рф/

I added `def compile_target_uri()` function which concatenates base URL and the query string. This funciton is called at the head of every single Flask `@app.route()` handler. All tests seem to pass.

It is better to handle this inside the Flask handler rather than do it inside `resolveMemento()`: the latter is a piece of core business logic and should not depend on specifics imposed by particular web framework, for example Flask.

I also updated the version of `ipfshttpclient` due to this (only yesterday resolved) bug: https://github.com/ipfs-shipyard/py-ipfs-http-client/pull/194


.gitignore
```diff
@@ -65,4 +66,4 @@ target/
 .ipynb_checkpoints
 
 # Mac meta
-.DS_Store
+.DS_Store
```
ibnesayeed, line 69: This removal of the trailing new line character is unnecessary. POSIX standard requires even the last line of a file to have a trailing new line character.

ipwb/replay.py
```diff
@@ -540,9 +553,6 @@ def generateCDXJTimeMapFromCDXJLines(cdxjLines, original, tmself, tgURI):
 
 # Fixes Flask issue of clipping queryString
 def getCompleteURI(uri):
```
ibnesayeed, line 555: Why is this `getCompleteURI` function which is now converted to a  do nothing type function left here if you were to define a replacement function `compile_target_uri`?

ipwb/replay.py
```diff
@@ -253,8 +249,22 @@ def resolveMemento(urir, datetime):
     return (newDatetime, linkHeader, uri)
 
 
+def compile_target_uri(url: str, query_string: bytes) -> str:
```
ibnesayeed, line 252: Is this `compile_target_uri` function any better, more accurate, or more efficient than previously defined `getCompleteURI`?

.gitignore
anatoly-scherbakov: Agreed, I think this is an artifact of my PyCharm or something. I am adding the newline with a new commit :)

ipwb/replay.py
anatoly-scherbakov: Oops, I forgot to remove it. Done now.

ipwb/replay.py
anatoly-scherbakov: 1. Missing query string is not a bug of Flask, contrary to the comment at `def getCompleteURI(uri)` - that's how Flask works. Thus, I moved the call of this function from the internals of `ipwb`'s business logic into Flask handlers. Business logic should not care about such things.

2. `getCompleteURI` is not a pure function since it relies upon global `flask.request` singleton. That makes it hard to test. `compile_target_uri` does not have that problem, and unit tests for it are in this PR.

3. I confess I could not resist the urge to change the piece of code I was modifying a bit towards Python's conventions. One of those conventions is:

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

That's why I renamed the function and, as a bonus, added type annotations.

I am also aware that all the other functions in the project are named in camelCase, more in Java tradition, and my change breaks consistency, which is bad. I can make a PR to rename all the other functions from camelCase to snake_case if you'd be open to that.

4. New function is a bit more verbose, but I also believe it is easier to understand.

machawk1: @anatoly-scherbakov Thanks for your contributions, the code needs more eyes for sanity checks by those more familiar with conventions (naming) and imported modules' base functionality (e.g., Flask).

I would be open to overhauling the naming if it puts us more inline with Python standards, which might in turn facilitate further contributions by those already familiar with the style.

ibnesayeed: Thanks @anatoly-scherbakov for putting forward your rationale. These camelCase function names are footprints of early code, which Mat was gradually changing over time. I think the most strong argument from you is the purity of the function which allows reliable unit testing.

Your `str.format()` calls are more verbose than they need to be. They are not bad, but I just find it way too verbose to to define non-reusable variables and assigning values to them and immediately using them. A string of joust a couple of pieces can perhaps be composed using positional placeholders without the fear of readability issues. Ideally, I would love to use f-strings for better readability, but that change should happen everywhere at once, which will also up the minimum Python version requirement to 3.6+.

Type annotations are great, but they would mean we are not supporting anything before Python 3.5, right? Also, should we defer type annotations to apply globally at once in a separate PR rather than introducing one function looking different from the rest?

anatoly-scherbakov: 1. A variable created once and once used is usually called an *explanatory variable*. Its point is to name a thing for the person, not for the computer, and show a human what is going on.

In this case, keyword arguments to `.format()` are indeed a bit too much - but I don't see they hurt, either :) I can remove them if you guys want more brevity.

> Type annotations are great, but they would mean we are not supporting anything before Python 3.5, right?

That's correct. However, I am not sure whether this is an issue since I am not sure in what environments this project is being used. It is not a library, so it can't break someone's project by not supporting an older Python version.

> Also, should we defer type annotations to apply globally at once in a separate PR rather than introducing one function looking different from the rest?

I do not believe so; there is a notion of "gradual typing" in Python world meaning a slow step by step addition of type hints into a Python project.

One of the reasons for this is that a PR to add all types everywhere AND make mypy happy would be too large. I prefer smaller PRs, easier to review and to think about.

anatoly-scherbakov: > I would be open to overhauling the naming if it puts us more inline with Python standards, which might in turn facilitate further contributions by those already familiar with the style.

If you create an issue, I will be happy to do something with it over this week. :)

ipwb/replay.py
```diff
@@ -252,9 +254,9 @@ def resolveMemento(urir, datetime):
 def compile_target_uri(url: str, query_string: bytes) -> str:
     """Append GET query string to the page path, to get full URI."""
     if query_string:
-        return '{path}?{query_string}'.format(
-            path=url,
-            query_string=query_string.decode('utf-8'),
+        return '{}?{}'.format(
+            url,
+            query_string.decode('utf-8'),
         )
```
ibnesayeed: With this reduced verbosity it might fit in a single line without complaining about the line length. Does it still have to be broken in multiple lines?

```suggestion
        return '{}?{}'.format(url, query_string.decode('utf-8'))
```

machawk1:
LGTM, too, merging.

Issue #634, opened 2020-04-18
Cannot fetch CDXJ file from IPFS


anatoly-scherbakov:
Running the following command on the current `master`:

    ipwb replay QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W

Returns this:

```
File QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W does not exist locally, fetching remote
No scheme in path, assuming IPFS hash and fetching...
Trying to ipfs.cat('QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W')
Data successfully obtained from IPFS
IPWB replay started on http://localhost:5000
 * Serving Flask app "ipwb.replay" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
```
Now, when opening http://localhost:5000, the server crashes:

```
Retrieving URI-Ms from QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W
File QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W does not exist locally, fetching remote
No scheme in path, assuming IPFS hash and fetching...
Trying to ipfs.cat('QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W')
Data successfully obtained from IPFS
a bytes-like object is required, not 'str'
(<class 'TypeError'>, TypeError("a bytes-like object is required, not 'str'"), <traceback object at 0x7f1de3dd4440>)
  File "/home/anatoly/.pyenv/versions/3.8.1/envs/ipwb/lib/python3.8/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/anatoly/.pyenv/versions/3.8.1/envs/ipwb/lib/python3.8/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/anatoly/.pyenv/versions/3.8.1/envs/ipwb/lib/python3.8/site-packages/ipwb/replay.py", line 593, in showLandingPage
    mementoInfo = calculateMementoInfoInIndex(iFile)
[...]

PR #635, opened 2020-04-18
#634 Fetch CDXJ files from IPFS

anatoly-scherbakov:
1. It appears that, in multiple cases, the application tries to fetch CDXJ file with plain `open()` call, assuming the file is local:

```
(ipwb) anatoly@anatoly-workstation:ipwb$ grep 'with open' ipwb/replay.py 
    with open(indexPath, 'r') as f:
    with open(cdxjFilePath, 'r') as f:
    with open(fullFilePath, 'r') as cdxjFile:
```

This obviously does not work with CDXJ files provided as IPFS hashes. To resolve this, I have reused `def getIndexFileContents()` in each of those cases. This function will handle determination of where to fetch the file from.

2. Additionally, I added `@lru_cache()` to cache the content of the file in memory and avoid fetching it separately for every request.

3. Plus, a bit of refactoring (`index.py` created for functions manipulating web archive index). Changes towards PEP8 compatibility and Python conventions.

After all these, this works for me now:

```
ipwb replay QmReQCtRpmEhdWZVLhoE3e8bqreD8G3avGpVfcLD7r4K6W
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ coverage.xml
 *,cover
 .hypothesis/
 tests/samples/warcs/frogTest_*
+samples/warcs/frogTest_*
 
 # Translations
 *.mo
--- a/ipwb/backends.py
+++ b/ipwb/backends.py
@@ -1,46 +1,40 @@
 import os
+from urllib.parse import urlparse
 
 import ipfshttpclient
-from ipfshttpclient.exceptions import StatusError
+import requests
 
 from ipwb import util
 
 
+def fetch_ipfs_index(path: str) -> str:
+    """Fetch CDXJ file content from IPFS by hash."""
+    with ipfshttpclient.connect(util.IPFSAPI_MUTLIADDRESS) as client:
+        return client.cat(path).decode('utf-8')
+
+
+def fetch_web_index(path: str) -> str:
+    """Fetch CDXJ file content from a URL."""
+    return requests.get(path).text
+
+
 def fetch_remote_index(path: str) -> str:
-    """Fetch CDXJ file from IPFS."""
-    path = path.replace('ipfs://', '')
-    # TODO: Take into account /ipfs/(hash), first check if this is correct fmt
+    """Fetch CDXJ file content from a remote location."""
 
-    if '://' not in path:  # isAIPFSHash
-        # TODO: Check if a valid IPFS hash
-        print('No scheme in path, assuming IPFS hash and fetching...')
-        try:
-            print("Trying to ipfs.cat('{0}')".format(path))
-            with ipfshttpclient.connect(util.IPFSAPI_MUTLIADDRESS) as client:
-                return client.cat(path).decode('utf-8')
+    if path.startswith('Qm'):
+        return fetch_ipfs_index(path)
 
-        except StatusError as err:
-            raise Exception((
-                'Cannot find CDXJ index file by hash {hash} in IPFS.'
-            ).format(
-                hash=path,
-            )) from err
+    scheme = urlparse(path).scheme
 
-    # TODO should be refactored into another function
-    else:  # http://, ftp://, smb://, file://
-        print('Path contains a scheme, fetching remote file...')
-        return util.fetch_remote_file(path)
+    if scheme == 'ipfs':
+        return fetch_ipfs_index(path.replace('ipfs://', ''))
 
-    # TODO: Check if valid CDXJ here before returning
+    elif scheme:
+        return fetch_web_index(path)
 
 
 def fetch_local_index(path: str) -> str:
     """Fetch CDXJ index contents from a file on local disk."""
-    # TODO what does this mean?
-    index_file_path = path.replace('ipwb.replay', 'ipwb')
-
-    print('getting index file at {0}'.format(index_file_path))
-
     with open(path, 'r') as f:
         return f.read()
 
--- a/ipwb/backends.py
+++ b/ipwb/backends.py
@@ -1,4 +1,3 @@
-import functools
 import os
 
 import ipfshttpclient
@@ -7,7 +6,7 @@ from ipfshttpclient.exceptions import StatusError
 from ipwb import util
 
 
-def _fetch_index_file_from_ipfs(path: str) -> str:
+def fetch_remote_index(path: str) -> str:
     """Fetch CDXJ file from IPFS."""
     path = path.replace('ipfs://', '')
     # TODO: Take into account /ipfs/(hash), first check if this is correct fmt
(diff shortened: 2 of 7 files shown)


anatoly-scherbakov:
Closes #634

ibnesayeed:
A couple of comments here:

* I do not think we should move the `fetchRemoteCDXJFile` function out of the replay.py to the index.py because that functionality is not needed during indexing, but only needed by the replay
* While it is important to cache index fetched from the IPFS as it is immutable, it might be problematic to cache the generic index loading function as it would prevent the ability to load live changes in the local index file unless the replay server is reloaded

anatoly-scherbakov:
1. The first motivation for moving the functions was that `replay.py` is more than 1100 lines long, which makes navigation and debugging rather difficult. It can be refactored into a bunch of smaller files, like this, for example:

```
index.py
memento.py
search.py
api.py
```

I agree that, right now, the name of `index.py` is confusing because there is already another file named `indexer.py`. My intent was to try factoring out functions that work directly with CDXJ indexes - in this case, they read an index file content from a local or remote location.

What is your view on refactoring these files?

2. I see; I did not consider the possible usecase of mutable `.cdxj` file. I can see it being a useful thing for live reload while testing and development. Since this is a testing/development use case, I suggest to use caching by default and turn it off using a command line option, say, `--no-cache`. 
When invoked with a HTTP/FTP/... index file source, this option should print a warning about degraded performance. When invoked with IPFS source, it should also say that disabling caching makes no sense with IPFS since content-addressed file is immutable.

What are your thoughts?

ibnesayeed:
> What is your view on refactoring these files?

I am not against refactoring. Me and @machawk1 spent countless hours talking about rearchitecting and refactoring the system and potentially rewriting from scratch after all the lessons we learned from this organically growing system. As I mentioned earlier, the project was hacked together in a hackathon in a limited amount of time, so at that time we only thought about to major components, an indexer and a replay. All the functions were organized in these two buckets, but later some common functions were moved to a shared util file. That said, there is no harm in rethinking it in terms of various classes of objects and organize components that way, then the indexer and replay would be nothing but two commands that compose these objects as they seem fit.

Since, mostly @machawk1 works on this codebase and he perhaps has a mental map of where things are and how theya re glued together, so I would ask him about it, what he feels. Should this kind of progressive refactoring be done now or an overhaul would be a better idea.

ibnesayeed:
> I can see it being a useful thing for live reload while testing and development.

On the contrary I think live reload is more desired in a production environment than in development (for this particular project). A production service will not be considered great if the service has to go down briefly each time more resources are crawled and indexed (which may happen rather frequently in busy archives). However, this can be achieved by introducing custom signals, they way many web servers work for log rotation and configuration reloading. I would personally go with opening the index file (if it is local) once, and use that file handle throughout the session, unless there is an interrupt to reload the configuration. This will avoid file opening and closing overhead each time. The current setup was a design decision @machawk1 took, perhaps motivated by some of his use cases or for simplicity in getting things to work.

And yes, remote resources should certainly be cached at the boot time of the replay server, irrespective of how we handle local index files. Ideally, we should be moving towards a CDX API driven approach that can independent necessary index functionalities independently using any back end as long as it has a consistent API, and one of those back ends could be memory, but we are not there yet.

PS: Me and Mat always fight for perfection versus getting things to work quicker. ;-)

anatoly-scherbakov:
> A production service will not be considered great if the service has to go down briefly each time more resources are crawled and indexed

Sure, but will that be that same CDXJ file? If we rely upon IPFS for our operations, I can only imagine that every new indexing of a website will be another CDXJ file.

Imagine a web server which will serve this distributed archive to the end users, something like, I don't know, global-archive.net. You will probably have to open global-archive.net/QmHash to open a given CDXJ file.

ibnesayeed:
> Sure, but will that be that same CDXJ file? If we rely upon IPFS for our operations, I can only imagine that every new indexing of a website will be another CDXJ file.

Supporting remote files as index was not a thing in the initial days, but @machawk1 added that at some point. Honestly speaking, I am not in favor of storing CDXJ index in IPFS to be accessed as an index file (there is no harm in storing it as regular file like others though), as archive index files are often mutable and storing mutable content in IPFS cuses unnecessary garbage over time which is not easy to clean (because IPFS does not have a built-in means of file history management). Other replay systems usually allow specifying a directory where CDX/CDXJ files are kept, which are all scanned at the lookup time (i.e., linear binary search). This setup allows incremental addition of index files and periodic merger for performance.

anatoly-scherbakov:
I've got multiple pains from mutable data, that's why I like immutability :) In my use case, I intend to create a separate CDXJ file every time I archive my web page, and put that file on IPFS. But as I replied in other thread, I will look into removing in-memory cache for now so that it can be reimplemented on another level, in a backend.

anatoly-scherbakov:
@machawk1 @ibnesayeed the cache removed and a bit of refactoring done.

I did not implement the proper disk cache though, I believe that would be a subject of another GitHub issue.

machawk1:
LGTM

ibnesayeed:
> In my use case, I intend to create a separate CDXJ file every time I archive my web page, and put that file on IPFS.

This approach will not scale for too long because while one can perform binary search in individual CDX files, looping through all the index files will be a linear operation. That's why there is a need of periodic merging. Lookup cost of K equal size sorted files with N lines each would be `K * log(N)` while after merging them the cost will be reduced to `log(K * N)`, which is a significant reduction if K is big.

machawk1:
@ibnesayeed Any objections to merging this PR?

ibnesayeed:
I don't think I have any major objections now other than the name of the file `backends.py` which is rather generic. The term backend alone does not tell what the file is all about. Should we go for something along the lines of `index_backend`?

ibnesayeed:
Thinking out loud, we should even call the base file as `index_backend`, but something like `index` that has generic functions, with more specific backend implementations in a sub-folder.

machawk1:
Naming is hard. :-)

I believe @anatoly-scherbakov proposed some potential revision to the py file naming schemes elsewhere. Backend seems somewhat descriptive of what the code in the file is doing without convoluting the code hierarchy. I move that we merge this and revisit revising the naming of both backend as well as indexer.py in a separate issue/PR.

Any objections to this action, @ibnesayeed?

ibnesayeed:
Sounds good to me!

Naming is hard you said? https://web.archive.org/web/20171224114025/https://www.thc.org/root/phun/unmaintain.html ;-)

machawk1:
Thanks again for your contribution, @anatoly-scherbakov!

Issue #639, opened 2020-04-26
ipwb index outputs unrelated information into CDXJ file


anatoly-scherbakov:
I am creating a CDXJ file from a WARC archive in a bash script, like this:

```shell_script
ipwb index "$warc_file.warc.gz" > "$warc_file.cdxj"
```

I subsequently put the file into IPFS, where it can be easily looked up, and an issue can be demonstrated:

```
$ ipfs cat QmZup5pBAJoXv49eszWZ3miJ9VMPnsKF3Ez2qRHXZ2pzoC | head -n 4
This version of ipwb is outdated. Please run pip install --upgrade ipwb.
* Latest version: 0.2020.4.24.1847
* Installed version: 0.2019.8.15.213
!context ["http://tools.ietf.org/html/rfc7089"]
```

First three lines should not be printed: due to them, my CDXJ file is now invalid. I suggest to use `logger.warning` for these messages instead of using `print()`. This way, the messages will be sent to stderr rather than to stdout. I aim to prepare a PR for that purpose shortly.

ibnesayeed:
Yes, this is a bug. The `util.py` file has a function defiled as `logError()` for this purpose, which should be used instead of `print()`.

anatoly-scherbakov:
1. I suggest to use Python's built-in `logging` instead a wrapper function around `print()`. There are plugins for flake8 which consider `print()` an antipattern.

2. I would actually like to remove the call of `util.check_pypi_for_update()` from `def main()`. This hurts performance; it slows down tests; if the `index` operation is performed on really many WARC files - there is a potential to DoS pypi.org. Maybe this check can be performed by a CLI subcommand, something like `ipywb check-for-update`?

machawk1:
I am unable to replicate this behavior and want to nail down the _why_ so we can also fix it in other places like you have provided in #640, @anatoly-scherbakov.

Perhaps it is the WARC itself but maybe it is the OS. I am using macOS 10.15.4 Python 3.8.0 and ipwb 0.2019.08.15.0213 from pypi (reinstalled previous version for testing).

```
% ipwb --version
InterPlanetary Wayback 0.2019.08.15.0213
% ipwb index samples/warcs/5mementos.warc > warc_file.cdxj
Processing WARC records in 5mementos.warc complete
% ipfs add warc_file.cdxj
added QmboaMdp1VJGqUfhYjva3JrQz3QbsJft4RVQ42sEp4tE3F warc_file.cdxj
[...]

PR #640, opened 2020-04-27
Issue 639 prints in cdxj

anatoly-scherbakov:


anatoly-scherbakov:
Closes #639

ipwb/__main__.py
```diff
@@ -1,26 +1,23 @@
...
-from . import util as ipwbUtil
-
-from .util import IPWBREPLAY_HOST, IPWBREPLAY_PORT
+from . import util
 
 
 def main():
-    ipwbUtil.checkForUpdate()
-    args = checkArgs(sys.argv)
+    util.check_pypi_for_update()
```
machawk1: At one point long ago, we did call this variable `util` but ran into some naming clash. It might be moot with the move to Py3 and refinement of some extraneous dependencies. @ibnesayeed?

ibnesayeed: Yes @machawk1 I recall you telling me about some conflicts long time ago when I asked why did you crate an alias for this import, but I think it should be alright now. Also, I think this import aliasing is done in other files too.

ipwb/check_for_update.py
```diff
@@ -0,0 +1,30 @@
...
+import requests
+
+from ipwb import __version__ as ipwb_version
+
+logger = logging.getLogger(__name__)
+
+
+def get_latest_ipwb_version() -> str:
+    """Fetch latest IPWB version from PyPI."""
+    response = requests.get('https://pypi.python.org/pypi/ipwb/json').json()
```
ibnesayeed, line 13: This may raise an exception if the PyPI service is not healthy, which should be handled.

ipwb/check_for_update.py
```diff
@@ -0,0 +1,30 @@
...
+    return response['info']['version']
+
+
+def check_pypi_for_update():
+    """Check if IPWB is outdated and print out a message if it is."""
+    latest = get_latest_ipwb_version()
+    current = re.sub(r'\.0+', '.', ipwb_version)
+
+    if current != latest and current is not None:
```
ibnesayeed, line 22: There is a higher chance of `latest` being `None` than `current` which is a local information and should be accessible if it is imported properly. I think the `None` check was performed due to the complications of relative imports in Py3 when a file is executed as a script in different scenarios. Also, I think check for `None` should be placed prior to any equality/inequality checks in general.

ipwb/check_for_update.py
anatoly-scherbakov: I will add a bit of user friendliness.

ipwb/check_for_update.py
anatoly-scherbakov: This is a piece of original code, but the point is valid.

ipwb/check_for_update.py
machawk1: Eventually we may want to decouple from pypi (e.g., binaries on IPFS) but moving this magic URI string out of the code might help with code quality, albeit, this should be the only place we use it.

ipwb/util.py
```diff
@@ -63,20 +56,18 @@ def isDaemonAlive(daemonMultiaddr=IPFSAPI_MUTLIADDRESS):
...
 
+    except OSError:
+        raise Exception(
+            'IPFS is likely not installed. See https://ipfs.io/docs/install/'
+        )
 
-def logError(errIn):
-    print(errIn, file=sys.stderr)
+    except Exception as e:
+        raise Exception('Unknown error in retrieving daemon status.') from e
```
machawk1, line 70: Will the removal of `logError(sys.exc_info()[0])` provide a sufficient traceback for debugging?

ipwb/util.py
```diff
@@ -220,10 +215,11 @@ def readIPFSConfig():
     try:
         with open(ipfsConfigPath, 'r') as f:
             return json.load(f)
+
     except IOError:
-        logError("IPFS config not found.")
-        logError("Have you installed ipfs and run ipfs init?")
-        sys.exit()
+        raise Exception(
```
machawk1, line 220: Same question as the above re:exceptions-in-exceptions and no longer exiting.

ibnesayeed:
Looks like Mat and I were looking at the same code at the same time. :smile:

ipwb/util.py
anatoly-scherbakov: Here is an example:
```pycon
In [1]: try: 
   ...:     raise ValueError('Ugly internal error, GRRRRR!!!') 
   ...: except ValueError as err: 
   ...:     raise Exception('Nice user-friendly error :)') from err 
   ...:                                                                                                           
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-8ed378f4cf49> in <module>
      1 try:
----> 2     raise ValueError('Ugly internal error, GRRRRR!!!')
      3 except ValueError as err:

ValueError: Ugly internal error, GRRRRR!!!

The above exception was the direct cause of the following exception:

Exception                                 Traceback (most recent call last)
<ipython-input-1-8ed378f4cf49> in <module>
      2     raise ValueError('Ugly internal error, GRRRRR!!!')
      3 except ValueError as err:
----> 4     raise Exception('Nice user-friendly error :)') from err
      5 

Exception: Nice user-friendly error :)
```

As you can see, the user will see the error message you supplied in `raise`, together with its stack trace. And if the user is to scroll a bit upper they will see the original exception properly preserved.

ipwb/util.py
```diff
@@ -63,20 +56,18 @@ def isDaemonAlive(daemonMultiaddr=IPFSAPI_MUTLIADDRESS):
...
-    except OSError:
-        logError("IPFS is likely not installed. "
-                 "See https://ipfs.io/docs/install/")
-        sys.exit()
-    except Exception as e:
-        logError('Unknown error in retrieving daemon status')
-        logError(sys.exc_info()[0])
 
+    except OSError:
+        raise Exception(
```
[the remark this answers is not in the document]
anatoly-scherbakov: The prior implementation exited; an unhandled Exception causes the program to do the same. So, no fundamental difference. But, using `raise ... from ...` prints the original stacktrace without any extra actions - just using the language's built-in mechanisms. That's why I prefer it to explicit `sys.exit()`.

What is the reason of re-raising the exception though? The same reason which probably motivated to catch these exceptions, in the first place - to provide a bit more friendly error message to the user, I believe.

The more proper way would be to create custom exception classes and to catch them at an upper level of the application converting to nice readable messages, but I believe this can be implemented a bit later.

anatoly-scherbakov:
I do not believe unhandled exceptions are inherently a bad thing. What I believe needs to be done is creating custom exceptions for various situations specific for the system, and handling them all in one place.

Something like this:

```
@dataclasses.dataclass(frozen=True)
class IndexNotFound(Exception):
    path: str
    backend: str

    def __str__(self):
        return f'The CDXJ index file {self.path} was not found by {self.backend} backend. Please check the correctness of the path.'
...

try:
    return replay(path)

except (IndexNotFound, InvalidIndex, ...) as err:
    if verbose is True:
        logger.exception()
    else:
        logger.warning(str(err))
    return
```

But this is tightly coupled with backends system because most of the operations with the index which can cause exceptions are probably going to be in the backend.

ipwb/util.py
machawk1: Ok, I am convinced. PEP 415's`raise exc from None` will also help suppress context, which is useful for development but noisy for the end user who should just be told directly what is wrong. I am fine using the pattern provided in this PR.

ibnesayeed:
I happy with raising custom exceptions if generic exception messages are too vague and a custom one can provide more context-specific message. However, I am against leaving known exceptions unhanded and let the program crash. There are situations where the program must exit/terminate on specific exceptions, but those exists should be preformed after catching those exceptions. There is a thin line between an application exit and application crash, unhanded exceptions suggest the latter. While stack trace is helpful for development, it is so not welcome by end users. Besides, if we know the cause of certain exceptions because we raised them, then we do not need to see the stack trace. Unknown issues should be the once that need to show stack trace and be handled after being discovered. I have no issues in having a top level exception handler, as long as it is specific enough to not mask previously unknown exceptions.

anatoly-scherbakov:
I will look into implementing some custom exceptions and their printouts in nearest days.

anatoly-scherbakov:
I again fell into the same pit of accumulating changes, so let me try to describe them in detail.

# Communicating errors using exceptions

I added a wrapper called `exception_handler()`, which is used as a decorator around `def main()`. It catches unhandled exceptions. When caught an exception, it does one of two things:

- if environment variable `DEBUG` is False (the default) - we will see only a logging message. Like this:

```shell
$ ipwb index ~/projects/covid-in-russia/warc/stopcoronavirus-russia.1588018596.cdxj
2020-07-05 23:51:43,336 [CRITICAL] ipwb: Daemon is not running at: /dns/localhost/tcp/5001/http
```

But we also can get full traceback:

```shell
$ DEBUG=true ipwb index ~/projects/covid-in-russia/warc/stopcoronavirus-russia.1588018596.cdxj
# --snip --
...exceeded with url: /api/v0/id?stream-channels=true (Caused by NewConnectionError('<ipfshttpclient.requests_wrapper.HTTPConnection object at 0x7f7bba80f4f0>: Failed to establish a new connection: [Errno 111] Connection refused'))

The above exception was the direct cause of the following exception:

# -- snip --

    raise Exception(f'Daemon is not running at: {daemonMultiaddr}') from err
Exception: Daemon is not running at: /dns/localhost/tcp/5001/http
```

The `DEBUG` variable also increases the logging level to ease debugging.

# IPFS client

After adding commits, codecov was unhappy about their coverage. I decided to try improving the situation by adding some tests, and when that didn't help - went on to try simplifying a few things. Fewer the lines, better the coverage.

`createIPFSClient()` function is changed to `ipfs_client()`. That function is doing what the original function did - it creates the client.

However, it also is wrapped in `@lru_cache` decorator which means that the value of the client will be cached. This removes the need for global `IPFS_API` variables - and thus I removed them.

...and only at that point did I realize that my goal of passing coverage barrier should have been achieved by just writing tests for `exception_handler()` - and I got to write them, which seems to finally make codecov happy.

anatoly-scherbakov:
@machawk1 fyi :)

ipwb/util.py
```diff
@@ -212,10 +226,11 @@ def readIPFSConfig():
...
             return json.load(f)
-    except IOError:
-        logError("IPFS config not found.")
-        logError("Have you installed ipfs and run ipfs init?")
-        sys.exit()
+
+    except IOError as err:
+        raise Exception(
+            'IPFS config not found. Have you installed ipfs and run ipfs init?'
+        ) from err
```
ibnesayeed, line 233: Since `sys.exit()` call is removed, and the wrapper exception handler will simply log the exception message when DEBUG is set to false, how will the process terminate? There is no point in keeping the service alive in an defined state when critical pieces are missing, no matter DEBUG is on or off. DEBUG should make logging more verbose with stack trace, but it should not change the process termination behavior. Current implementation, if I could understand it well, means it will terminate the process by raising the exception when DEBUG is on even when the exception is rather innocent and handlable and will not terminate the process even on critical exceptions when the DEBUG is off. Is this the case or am I missing something here?

anatoly-scherbakov: @ibnesayeed let us describe the calls sequence:

```python
if __name__ == "__main__":
    main()
```

seemingly calls `main()`, but _actually_ calls `wrapper(*args, **kwargs)` (see the implementation of `exception_logger`).

Then, `wrapper` calls the original `main()` inside a `try - except` block.

Now what happens if an unhandled exception pops out somewhere in `main()` or one of the functions it calls? Being unhandled, that exception pops up through the call stack, reaches `main()`, passes it and arrives at `wrapper()`.

The execution of `main()` at this point is already interrupted by a propagating exception. `wrapper()` now can either reraise the exception or just log it - that does not matter because, when its `except` clause is done its work, `wrapper()` function returns and the process stops.

At least it should if everything works as I believe it does.

ibnesayeed: @anatoly-scherbakov I understand how wrapper/decorator functions work in Python. I have actually used them heavily in [an HTTP testing framework](https://github.com/ibnesayeed/webserver-tester) I wrote a while ago. My concern here is a little different. Not every exception needs to be reraised to bubble out of the main and let the process die. For example, if the system is configured to use a local IPFS node (which is the case for now as the code needs some changes to decouple IPFS from IPWB) and the IPFS daemon is not accessible because the binary is not installed on the system then the process should die because there is no hope it will ever be accessible. However, if the daemon is not accessible because it is not running, but is installed, then we may not want to let the process die, because the admin control panel has an option to attempt to start/stop IPFS daemon. In the earlier code, some exceptions were logging a message and returning `None` while other exceptions were logging a message and exiting. I am not sure if that behavior is still preserved. I think it is great that we have a catch-all wrapper exception handler, but we may want to handle some exceptions inline and take appropriate actions instead of bubbling every exception up with a custom message.

ibnesayeed: Never mind, I think the custom exception `IPFSDaemonNotAvailable` is handled inline in the replay. This looks good to me now. Thanks @anatoly-scherbakov for your valuable contributions. We look forward to see more from you.

machawk1:
Did some manual verification on my end as well. Moving to merge this into master. Thanks, @anatoly-scherbakov!

Issue #638, opened 2020-04-25
Issues when using go-ipfs 0.5.0


machawk1:
In anticipation of the 0.5.0 release of go-ipfs, I tested the current master (ipwb v0.2020.04.24.1847 release) against the latest prerelease of ipfs (ipfs version 0.5.0-rc3) with latest ipfshttpclient (ipfshttpclient==0.4.13.2).

In a virtualenv with Python 3.7/3.8:

```bash
> ipwb index samples/warcs/5mementos.warc
Unknown error in retrieving daemon status
<class 'ipfshttpclient.exceptions.StatusError'>
```

In its README, ipfshttpclient (nee the ipfsapi Python module) states it is tested against go-ipfs v0.5.0-rc2. Repeating the above with this version produces the same error.

machawk1:
Testing against latest RC (0.5.0-rc4) seems to be attribute to a change in ipfshttpclient:

```py
import ipfshttpclient
x = ipfshttpclient.Client('/dns/localhost/tcp/5001/http')
x.id()
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/ipfshttpclient/http.py", line 266, in _do_raise_for_status
    response.raise_for_status()
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 405 Client Error: Method Not Allowed for url: http://localhost:5001/api/v0/id?stream-channels=true

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/ipfshttpclient/client/base.py", line 15, in wrapper
    result = func(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/ipfshttpclient/client/miscellaneous.py", line 89, in id
    return self._client.request('/id', args, decoder='json', **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/ipfshttpclient/http.py", line 51, in wrapper
    return func(self, *args, **merged)
[...]

PR #641, opened 2020-04-28
Temporarily install ipfshttpclient from master branch

ibnesayeed:
A recent change in the IPFS API removed support for GET/HEAD methods on some endpoints. This prevents us from upgrading to go-ipfs-v0.5.0 as per #638. Corresponding change in the `ipfshttpclient` is added to the `master` branch, but not released on PyPI yet. This is a temporary fix, but once the updated package is out, we should update the dependency.

Since this is a temporary fix, I am not changing the dependency in the `setup.py` file. Should I?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 warcio>=1.5.3
-ipfshttpclient>=0.4.13
+git+git://github.com/ipfs-shipyard/py-ipfs-http-client@master#egg=ipfshttpclient
 Flask==1.1.1
 pycryptodome>=3.4.11
 requests>=2.19.1
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     ],
     install_requires=[
         'warcio>=1.5.3',
-        'ipfshttpclient @ git+https://github.com/ipfs-shipyard/py-ipfs-http-client@master',
+        'ipfshttpclient @ git+https://github.com/ipfs-shipyard/py-ipfs-http-client@master',  # noqa: E501
         'Flask==1.1.1',
         'pycryptodome>=3.4.11',
         'requests>=2.19.1',
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     ],
     install_requires=[
         'warcio>=1.5.3',
-        'ipfshttpclient>=0.4.12',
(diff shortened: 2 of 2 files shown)


ibnesayeed:
I made changes in the setup.py file as well, but now it is complaining about lne too long, thought I would not like to break URLs to pass the linter.

ibnesayeed:
Thanks for this @machawk1, yesterday I was looking for annotations to add exception here, but I did not search well enough. From the brief look at the documentation I thought they have options to disable certain validations globally, but not per line exceptions.

machawk1:
@ibnesayeed LGTM as a temp fix. I added a directive to the setup file in 070ef18 to allow the line to pass the linter. In an effort to stay current, should we push another release for compatibility's sake?

ibnesayeed:
> In an effort to stay current, should we push another release for compatibility's sake?

Let's wait for the new release of the `ipfshttpclient` before we push a new release. It is our primary dependency and I do not feel comfortable tagging a release that has non-frozen dependencies. Also, I am not too confident that PyPI will install the dependency properly when releasing there because there have been some changes in how dependencies are specified from a remote resource  in the `setup.py` file. However, if the `ipfshttpclient` takes forever to release the new version, we may revisit our plan after a while.

machawk1:
Ok, I agree, but the latest release of ipwb does not work with the latest release of go-ipfs currently. Do you want to revisit this in a few days, @ibnesayeed, and push a new ipwb release with a reference to the ipfshttpclient master if ipfshttpclient has not released a new version by then? Should we pin to an ipfshttpclient commit instead of simply master? Pinning to master seems dangerous and potentially breaking.

ibnesayeed:
If we were to release in a few days, pinning to a hash will be safer, but not all intermediary references are fetchable from Git servers (a configuration for efficiency). We can perhaps ask the `ipfshttpclient` maintainer about their next release plan.

machawk1:
@ibnesayeed Can you take the lead on the task you mentioned? Feel free to reference the relevant GitHub issues in the ipwb repo.

ibnesayeed:
> Can you take the lead on the task you mentioned?

Posted a comment at https://github.com/ipfs-shipyard/py-ipfs-http-client/issues/204#issuecomment-621312577 to inquire about it.

PR #647, opened 2020-05-12
Update TravisCI to use go-ipfs 0.5.1

machawk1:
@ibnesayeed Can you verify that only these two files need to be changed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 python:
   - "3.7"
 before_script:
-  - wget "https://dist.ipfs.io/go-ipfs/v0.5.0/go-ipfs_v0.5.0_linux-amd64.tar.gz" -O /tmp/ipfs.tar.gz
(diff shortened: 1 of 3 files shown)


ibnesayeed:
Yes, generally that's all we need to change for IPFS version bump. However, it looks like the pycodestyle is not happy with the use of `l` as the variable in (which has nothing to do with this issue):

https://github.com/oduwsdl/ipwb/blob/b8c1ea350618ab1bdf6abba67f99316303f53ce9/ipwb/indexer.py#L151

machawk1:
Odd that this surfaced with this PR, maybe a change in pycodestyle. I will make the change, update this PR, then give you a ping before merging.

machawk1:
Yep, I updated my local pycodestyle from 2.5.0 to 2.6.0 and sure enough the same linting issue surfaced.

ibnesayeed:
Yes, that was my guess too, it was something newly introduced to the linter.

Issue #630, opened 2020-01-31
Consider moving to GitHub Workflow from TravisCI 


machawk1:
In #624, @ibnesayeed suggested we moved to using GitHub Workflow. We currently use TravisCI but no build matrix. I have used a build matrix with WAIL ([for example](https://github.com/machawk1/wail/blob/osagnostic/.travis.yml)) and @ibnesayeed has with [access-parser](https://github.com/oduwsdl/accesslog-parser/blob/master/.github/workflows/test.yml).

- Q: What (dis-)advantages does GitHub Workflow use compared to TravisCI?
- Q: Are there any barriers in using both? Are there any reasons not to do so?

- [ ] Initial task: move to using a matrix build in TravisCI (current master).

ibnesayeed:
In the accesslog-parser repo we use both TravisCI and GH Workflow, but planning on retiring the former which was put in place before I included the latter.

### Potential Advantages of Using Both Together

* The same code is tested in multiple places with different infrastructure backbone, sop platform-specific failures might surface up, though containerized testing environments mask some of these benefits out
* The code is distributes in more places, potentially, gaining more visibility
* Identifying a failure as transient or systemic is easier as the latter will more likely fail in both the places

### Potential Disadvantages of Using Both Together

* CI/CD pipeline is stalled as per the slowest channel of the two platforms (i.e., if GH performs tests in 30 seconds and TravisCI takes 5 minutes to pull the code and run tests due to network slowdown or overloaded queue then the next step in the pipeline has to wait for more than 5 minutes)
* Probability of transient failures increase with the increase of number or places where tests are performed, requiring manual interventions more often
* Inter-service access using access token increases the attack surface of credential leakage, especially when third-party services ask for more permissions than they need to function properly (or access permissions are not fine-grained enough)

machawk1:
@ibnesayeed Can you comment on how the two offerings (TravisCI and GH workflow) compare in the context for how we would anticipate to further use the services?

ibnesayeed:
Perhaps we can add GH Workflow for testing (and in the future for other tasks like publishing to PyPi) while keeping TravisCI in place. After a while if we feel that GH's in-house CI is serving the purpose well and the external service is an unnecessary overhead then we can retire that, otherwise keep that running too. Additionally, we may also incorporate Docker image building and publishing here in GH itself.

machawk1:
@ibnesayeed 👍 I have updated the original post on this GH issue to reflect these tasks.

[...]

PR #648, opened 2020-05-12
Add GH Workflow for cross-platform tests

ibnesayeed:
This utilizes matrix of OSes, Python and IPFS versions to perform tests of various combinations.


ibnesayeed:
It looks like Windows tests are halting. We need to test it manually on a Windows machine as following:

* Install Windows Server 2019
* Ensure Bash and cURL are installed in it
* Install NodeJS 12
* Install Python 3.7 or 3.8
* Install IPFS (i.e., download and extract the binary and set PATH accordingly)
* Run IPFS daemon
* Run ipfs cat on the welcome file
* Clone IPWB
* Run IPWB tests

ibnesayeed:
This aims to resolve #630, but Windows environment is not behaving well. I think we can move with what we have working here and revisit Windows later.

ibnesayeed:
> @machawk1: 3.8 is problematic on some systems

That's why we are testing 3.7 and 3.8 both on various platforms using matrix, but inside a Docker container we can more confidently go for the latest Python.

ibnesayeed:
@machawk1 should we abstract out the task of linting the code as a separate job and not repeat it in every matrix environment?

machawk1:
@ibnesayeed Yes, only need to lint in one environment. No need to repeat.

ibnesayeed:
@machawk1 this is in good shape (other than Windows, which is commented out for now).

machawk1:
@ibnesayeed I should have some time to explore and evaluate this tomorrow (May 14) afternoon.

ibnesayeed:
It looks like the `codecov` utility is a not an independent tool, but a way to submit results to codecov.io for reporting, hence removing it from the workflow.

ibnesayeed:
Windows tests are still failing and taking forever to complete, likely due to IPWB bugs (because the action otherwise runs tests on all platforms fine).

Issue #649, opened 2020-05-14
Revisit linting, testing, code coverage, and package publishing plans


ibnesayeed:
Over the time we have tried many different tools, packages, and services for linting, testing, and code coverage etc. As a result there are some files, dependencies, and configurations who's presence is unclear or confusing.

### Linting

* Do we want to continue using existing Python and JavaScript linters or should we update any of that?

### Testing

* Which testing framework do we want to use and what configurations do we want when running those tests?
* Remove any unnecessary files and dependencies and update dependencies to their latest versions.
* Which CI platforms do we want to use: Travis CI, GH Workflows, or both?

### Code Coverage

* Do we want to run local tools for code coverage, a service, or both?
* If we do choose to use a service, do we want to use their GitHub App (with more access) or incorporate their GH Action?

### Publishing

* Should we use GH Workflows to build and push releases to PyPI or let PyPI pull our code and package them on their servers (this has consequences of the origin of credentials/access tokens)?
* Should we let DockerHub auto-build images by pulling our code (which is taking longer these days and the automation is broken and it requires access to an entire GH account) or let images be built in GH Workflows and push them to DockerHub (this has consequences of the origin of credentials/access tokens)?
* Do we also want to release packages in [GH Package Registry](https://github.com/features/packages)?

PR #654, opened 2020-05-18
Utilize GH Workflow to publish to PyPI

ibnesayeed:
This implements the last point of #630. Also, related to #649.


ibnesayeed:
> Will this keep the releases on GitHub and pypi in step?

Releases on GH will be performed the same way as before (i.e., by running `release.sh` file locally). Installing `twine`, packaging, and publishing to PyPI are the steps that are taken out of the shell script and moved to this workflow. The `dist.yaml` workflow file utilizes a secret key that I have generated in PyPI and stored it under `Settings > Secrets` here.

machawk1:
Ok, this process should probably be documented. It's starting to sound a little Rube Goldberg-esque.

ibnesayeed:
I don't think we need to document something that is of no interest of users. For maintainers of a repo the process is being widely adopted by the GH community that it will become obvious how things are done, especially, these YAML files are rather self-descriptive. I feel like documenting it is similar to if every repo using a Dockerfile were to document what Docker is, how it works, and how to install it. If one has used any CI system before, things will be easy to follow and GH has great documentation on how various pieces work.

machawk1:
It is not for the users, @ibnesayeed, it is for me. Despite what is being adopted, "future us" would like to know. He has an axe.

ibnesayeed:
Fair enough! Which steps of the process do you think are not very straight-forward and you need to document them somewhere for future reference? I can try to explain.

machawk1:
A wiki page would be good that we could link from the README. I think I understand the process but will likely not look back on this GH issue (or PR!), so...adapt the above, in your words of expertise, to an external place.

ibnesayeed:
I have added some inline documentation in some obvious places where I think you might look when you need some clarification.

PR #662, opened 2020-06-18
Update ipfshttpclient dependency to ipfshttpclient4ipwb

machawk1:
This is due to ipfshttpclient not pushing a new release to pypi and ipwb needing to be compatible with newer go-ipfs binaries, which the currently released (to pypi) version of ipfshttpclient is not.

### DIFF
--- a/ipwb/backends.py
+++ b/ipwb/backends.py
@@ -2,7 +2,7 @@ import dataclasses
 from typing import Optional
 from urllib.parse import urlparse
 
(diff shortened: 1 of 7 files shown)


machawk1:
I agree. This is mainly applicable to {test-}backends.py, as the other scripts seem to have aliases in-place. I think the simple change in this PR will make reverting the name easy going forward when ipfshttpclient pushes another release to pypi.

Issue #283, opened 2017-11-22
Support Datetime other than those specified as 14-digits


machawk1:
The [WARC 1.1 spec](https://iipc.github.io/warc-specifications/specifications/warc-format/warc-1.1/) allows for more precise datetimes. These should be supported in the replay system. Does any tool exist that will generate these yet? If not, some sample data can be fabricated.

machawk1:
The further precision does not be present in the link above. What's the BnF link?

Also see https://github.com/iipc/warc-specifications/pull/21.

machawk1:
First line `WARC/1.1` causes an exception in the iterator we currently reuse from pywb to quickly invalidate the WARC and not proceed with processing.

machawk1:
Per the WARC/1.1 spec and https://github.com/iipc/warc-specifications/pull/21, date strings like "https://github.com/iipc/warc-specifications/pull/21" are legal but currently breaks the indexer with:

```py
Traceback (most recent call last):
  File "/usr/local/bin/ipwb", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/ipwb/__main__.py", line 17, in main
    args = checkArgs(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/ipwb/__main__.py", line 151, in checkArgs
    results.func(results)
  File "/usr/local/lib/python2.7/site-packages/ipwb/__main__.py", line 32, in checkArgs_index
    debug=args.debug)
  File "/usr/local/lib/python2.7/site-packages/ipwb/indexer.py", line 141, in indexFileAt
    warcFileFullPath, **encryptionAndCompressionSetting)
  File "/usr/local/lib/python2.7/site-packages/ipwb/indexer.py", line 179, in getCDXJLinesFromFile
    for i in iterForCounting(fhForCounting):
  File "/usr/local/lib/python2.7/site-packages/pywb/warc/archiveiterator.py", line 543, in __call__
    for entry in entry_iter:
  File "/usr/local/lib/python2.7/site-packages/pywb/warc/archiveiterator.py", line 379, in create_record_iter
    entry = self.parse_warc_record(record)
  File "/usr/local/lib/python2.7/site-packages/pywb/warc/archiveiterator.py", line 465, in parse_warc_record
    get_header('WARC-Date'))
  File "/usr/local/lib/python2.7/site-packages/pywb/utils/timeutils.py", line 122, in iso_date_to_timestamp
[...]

PR #663, opened 2020-06-20
Add test to show breakage using variable len WARC-Dates for #283

machawk1:
Merges new test re:#283 to show that the indexer does not throw an exception when trying to parse WARC/1.1 dates beyond the WARC/1.0 spec. This could be more sophisticated but is a start to integrating the variableSizedDate WARC into the ipwb test suite.

### DIFF
--- a/tests/test_indexing.py
+++ b/tests/test_indexing.py
@@ -32,9 +32,9 @@ def test_warc_ipwbIndexerBrokenWARCRecord():
 # exhibiting these dates.
 @pytest.mark.ipwbIndexerVariableSizedDates
 def test_warc_ipwbIndexerVariableSizedDates():
-    pathOfBrokenWARC = os.path.join(
-        os.path.dirname(__file__) +
-        '/../samples/warcs/variableSizedDates.warc')
+    pathOfBrokenWARC = \
+        os.path.normpath(os.path.join(os.path.dirname(__file__), '..',
+                         'samples', 'warcs', 'variableSizedDates.warc'))
     indexer.indexFileAt(pathOfBrokenWARC, quiet=True)
 
 # TODO: Have unit tests for each function in indexer.py
--- a/tests/test_indexing.py
+++ b/tests/test_indexing.py
@@ -27,4 +27,14 @@ def test_warc_ipwbIndexerBrokenWARCRecord():
     assert ipwbTest.countCDXJEntries(cdxj) == 1
 
 
+# WARC/1.1 allows for dates of length that are not easily converted
+# to 14-digits. This test highlights the failures from a WARC
+# exhibiting these dates.
+@pytest.mark.ipwbIndexerVariableSizedDates
+def test_warc_ipwbIndexerVariableSizedDates():
+    pathOfBrokenWARC = os.path.join(
+        os.path.dirname(__file__) +
+        '/../samples/warcs/variableSizedDates.warc')
+    indexer.indexFileAt(pathOfBrokenWARC, quiet=True)
+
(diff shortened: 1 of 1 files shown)


tests/test_indexing.py
```diff
@@ -27,4 +27,14 @@ def test_warc_ipwbIndexerBrokenWARCRecord():
     assert ipwbTest.countCDXJEntries(cdxj) == 1
 
 
+# WARC/1.1 allows for dates of length that are not easily converted
+# to 14-digits. This test highlights the failures from a WARC
+# exhibiting these dates.
+@pytest.mark.ipwbIndexerVariableSizedDates
+def test_warc_ipwbIndexerVariableSizedDates():
```
ibnesayeed, line 34: Why mixed-case function name?

tests/test_indexing.py
```diff
@@ -27,4 +27,14 @@ def test_warc_ipwbIndexerBrokenWARCRecord():
...
+    pathOfBrokenWARC = os.path.join(
+        os.path.dirname(__file__) +
+        '/../samples/warcs/variableSizedDates.warc')
```
ibnesayeed, line 37: When using `os.path.join`, each path component should be passed as individual parameters, so that the function uses an OS-dependent path separator.

tests/test_indexing.py
machawk1: Aligning naming to other tests. This PR is not about revising the naming scheme but to add another test.

tests/test_indexing.py
machawk1: This was copied from other tests. If we revise a means, we should do so across the board. I agree with your suggestion but currently tests are very sparse.

tests/test_indexing.py
ibnesayeed: Do you anticipate a separate PR fixing all the symbol names across the board?

tests/test_indexing.py
ibnesayeed: You may want to take this opportunity to correct it.

machawk1: Do you know of a standard approach at specifying this? I can imagine an assortment of ways. One in particular is to read the delimiter and `delimited.join()`.

tests/test_indexing.py
machawk1: It is something I would like to do -- align with Pythonic syntactic and semantic conventions after being more informed by @anatoly-scherbakov's contributions. This might be best described as a GitHub issue highlighting where the codebase strays away and making a TODO of fixes for conformance (e.g., to PEPs).

tests/test_indexing.py
ibnesayeed, line 37:
```suggestion
    pathOfBrokenWARC = \
        os.path.normpath(os.path.join(os.path.dirname(__file__), '..',
                         'samples', 'warcs', 'variableSizedDates.warc'))
```

ibnesayeed: I have added a code suggestion in a separate comment to tackle this.

tests/test_indexing.py
ibnesayeed: When I commented on it, I too was thinking about @anatoly-scherbakov's argument about fixing things as we encounter them rather than waiting for one big overhaul.

machawk1: Either approach is good, so long as we have a concrete target. My style is affected by conventions of a few past languages, but I am more than willing to adhere. It might be good to have some style check in-place to enforce this. Ideally, it would just fix it like gofmt.

PR #670, opened 2020-06-24
Update IPFS version to 0.6.0

### LINKED ISSUE #669: Update tests to use go-ipfs 0.6.0
It looks like a new version of [go-ipfs (v0.6.0)](https://github.com/ipfs/go-ipfs/releases/tag/v0.6.0) was released last week. I think we should update tests and builds to include that. Since the minor point release was bumped up rather quickly, I think we should test against the latest releases of the last three minor versions instead of just two.

ibnesayeed:
In addition to usual IPFS version bump, this PR also adds increases one of the matrix dimensions and closes #669.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,7 @@ jobs:
         ipfs:
           - 0.4
           - 0.5
+          - 0.6
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os }} Py-${{ matrix.python }} IPFS-${{ matrix.ipfs }}
     steps:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: python
 python:
   - "3.7"
 before_script:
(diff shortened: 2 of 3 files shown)


ibnesayeed:
The client needs to declare support for the new minor version. I have tried to make this change in https://github.com/machawk1/py-ipfs-http-client/pull/1, but not sure what else is required to bump the version support in the client.

machawk1:
@ibnesayeed I merged the PR in the py-ipfs-http-client repo as well as updated a couple other references. This has also been pushed to pypi and versioned to align with the go-ipfs client, as documented in the module's versioning scheme basis.

machawk1:
I manually re-triggered the TravisCI build following the push of the new version of the supporting module to pypi and the status changed to success. @ibnesayeed Can you do a final sanity check before we merge this to ipwb master?

ibnesayeed:
I re-ran all the workflows and they all seem to be passing. So, we should be good to go on this one.

PR #672, opened 2020-06-24
Mv string formatting in indexer to use f-strings

machawk1:


ipwb/indexer.py
```diff
@@ -272,7 +271,7 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
                     (hstr, payload, nonce) = \
                         encrypt(hstr, payload, encryptionKey)
 
-            # print('Adding {0} to IPFS'.format(entry.get('url')))
+            # print('Adding {entry.get('url')})
```
ibnesayeed, line 274:
```suggestion
            # print(f'Adding {entry.get("url")} to IPFS')
```

ipwb/indexer.py
```diff
@@ -306,19 +304,18 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
...
             cdxjLines.append(cdxjLine)  # + '\n'
         return cdxjLines
 
 
 def generateCDXJMetadata(cdxjLines=None):
     metadata = ['!context ["http://tools.ietf.org/html/rfc7089"]']
     metaVals = {
-        'generator': "InterPlanetary Wayback v.{0}".format(ipwbVersion),
-        'created_at': '{0}'.format(datetime.datetime.now().isoformat())
+        'generator': f'InterPlanetary Wayback v.{ipwbVersion}',
```
ibnesayeed, line 315: Should we have a dot after `v`?

```suggestion
        'generator': f'InterPlanetary Wayback v{ipwbVersion}',
```

ipwb/util.py
```diff
@@ -204,9 +204,9 @@ def fetch_remote_file(path):
         r = requests.get(path)
         return r.text
     except ConnectionError:
-        logError('File at {0} is unavailable.'.format(path))
+        logError(f'File at {path} is unavailable.')
     except Exception as E:
-        logError('An unknown error occurred trying to fetch {0}'.format(path))
+        logError(f'An unknown error occurred trying to fetch {path}')
```
ibnesayeed, line 209:
```suggestion
        logError(f'An unknown error occurred while trying to fetch {path}')
```

ipwb/indexer.py
machawk1: Oof, good catch! A second set of eyes is definitely helping here. :-)

ibnesayeed: If it was not a commented out line, you would have caught it in the first place I guess.

ipwb/indexer.py
machawk1: Yes, I think so, but am not tied to it. Should we even have a "v" and not just state "InterPlanetary Wayback 0.2020.06.24" in the indexes?

This also makes me wonder if we need some sort of regression/adaptation to indexes created with older versions, e.g., the metadata first character change.

ibnesayeed: Either way is fine, whether or not we include `v`, but we need to be consistent about it.

We do need to revisit old index files and adjust every now and then as things evolve.

machawk1: `ipwb --version` returns the version number appended without a "v". I believe the above code is the only instance where we generate new data with an ipwb version number. I would prefer removing the "v." altogether.

machawk1: I was wrong. We still use it in GitHub releases.

release.sh
```diff
@@ -16,13 +16,13 @@ git add 'ipwb/'$FILE_NAME
 git commit -m "RELEASE: Bump version to "$VERSION_STRING
 
 # Create a tag in repo
-TAG_NAME='v'$VERSION_STRING
+TAG_NAME='Release '$VERSION_STRING
 git tag $TAG_NAME
 git push
 git push origin $TAG_NAME
```
ibnesayeed, line 22: We don't want spaces in a tag name:

```suggestion
git tag $VERSION_STRING
git push
git push origin $VERSION_STRING
```

ipwb/indexer.py
```diff
@@ -306,19 +304,18 @@ def getCDXJLinesFromFile(warcPath, **encCompOpts):
...
         return cdxjLines
 
 
 def generateCDXJMetadata(cdxjLines=None):
     metadata = ['!context ["http://tools.ietf.org/html/rfc7089"]']
     metaVals = {
-        'generator': "InterPlanetary Wayback v.{0}".format(ipwbVersion),
-        'created_at': '{0}'.format(datetime.datetime.now().isoformat())
+        'generator': f'InterPlanetary Wayback {ipwbVersion}',
+        'created_at': f'{datetime.datetime.now().isoformat()}'
```
ibnesayeed, line 316: I think the `isoformat()` already returns a string, so no need to perform an unnecessary interpolation.

```suggestion
        'created_at': datetime.datetime.now().isoformat()
```

PR #675, opened 2020-06-28
WIP: GH Actions for Windows test

machawk1:
Check to see with the updated path conversion where we rely on the delimiter of the OS was an issue on Windows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
-          # - windows-latest
+          - windows-latest
         python:
           - 3.7
           - 3.8


ibnesayeed:
This needs to be tested on a Windows machine locally first to investigate what is wrong.

machawk1:
Why "first", @ibnesayeed? I thought it would be worthwhile to see how the change affected the test results.

ibnesayeed:
Because it looks like the STDOUT of tests is not being flushed and streamed on the web-based log viewer for Windows instances as they happen. This means, if the test is hung up for some reason, we will not know the outcome until the maximum timeout period is exhausted, which is several hours. This will not only eat up the computation quota, if there is any, but also make the turn around time really long as we iterate through changes to get things to work.

If we could investigate the issue locally fist and things start to work, then keeping Windows in the matrix will ensure its ongoing compatibility.

machawk1:
@ibnesayeed Thanks for the clarification. I should have a Windows environment available some time soon (for some [WAIL](https://github.com/machawk1/wail) testing), so can run run the ipwb tests in that environment and try to debug the output. Closing this PR now, as it is evident that the Windows testing is still running for longer than expected. I have also requested for the workflow run to be cancelled.

PR #686, opened 2020-07-02
Moves online version check to separate function.

### LINKED ISSUE #676: Remove pypi.org query when checking the version of ipwb
In #639, @anatoly-scherbakov highlighted that ipwb sends a request to pypi.org when the version flag is set. This flag does not traditionally hit the network and should return quickly. In the same GitHub issue, there seemed to be agreement that this change is needed but it was neither documented separately as a TODO or executed. The code remains in util.py.

This ticket serves as the TODO for this change to be made.

machawk1:
By default, the ipwb version check makes a request to pypi.org to determine if the
currently installed version is current. This should not occur -- the command should
be executed locally and the current version reported. However, this functionality is
useful, so I have put it behind a separate flag.

Closes #676

### DIFF
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -147,8 +147,7 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-u', '--update-check',
         action='store_true',
-        help=('Checks remotely to determine if the installed version '
-              'of ipwb is current')
+        help='Check whether an updated version of ipwb is available'
         )
     parser.set_defaults(func=ipwbUtil.checkForUpdate)
 
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -15,8 +15,7 @@ from .util import IPWBREPLAY_HOST, IPWBREPLAY_PORT
 
 
 def main():
-    ipwbUtil.checkForUpdate()
-    args = checkArgs(sys.argv)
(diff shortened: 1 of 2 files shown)


machawk1:
As a caveat, here, I had an issue getting the argparser to consume the command without a parameter or introducing a new base command like `ipwb index` and `ipwb replay`. I believe this solution is less verbose than a solution not resorting to setting the "default". I am open to another solution that works, but this fulfills the issue in #676.

ipwb/__main__.py
```diff
@@ -145,10 +144,18 @@ def checkArgs(argsIn):
     parser.add_argument(
         '-v', '--version', help='Report the version of ipwb', action='version',
         version='InterPlanetary Wayback ' + ipwbVersion)
+    parser.add_argument(
+        '-u', '--update-check',
+        action='store_true',
+        help=('Checks remotely to determine if the installed version '
+              'of ipwb is current')
```
ibnesayeed, line 151: I think there is no need to say that it will be done remotely, which is obvious.

```suggestion
        help='Check if an updated version of ipwb is available'
```

machawk1:
Adapted it further in c574416, @ibnesayeed. I think the "Check if" might unintentionally introduce a perceived conditional. The check is performed regardless of whether the version is current. "Whether" was substituted for "if".

PR #690, opened 2020-07-09
Cleanup logo source, add version with stroke for dark mode adaptation

### LINKED ISSUE #689: README logo is difficult to see in dark mode
I noticed that in the GitHub Android app that the project logo is difficult to distinguish from the dark, default background:

app_logo

This is also likely the case anywhere the background is dark. Per a discussion with @ibnesayeed, one solution is to provide a white stroke to all components of the logo.

machawk1:
This PR updates the PSD by removing a stray pixel from the "planet" part of the image source and provides a relatively sized stroke to all elements to enhance the appearance when in dark mode.

This PR does _not_ address the need to eventually use a vector format.

Closes #689


machawk1:
Still pending @ibnesayeed's review but:

- [ ] Update ipwb/assets/logo.png so the white-stroked logo is also present in the replay web interface.

ibnesayeed:
LGTM!

When adding in the assets, make sure you have it in the right size as we may not need a full-size image in the replay interface.

machawk1:
> When adding in the assets, make sure you have it in the right size as we may not need a full-size image in the replay interface.

Discussion of this in #689. I think the task is beyond the scope of the issue, as the replay web interface does not yet adapt to dark mode. It _should_, but that will require more thought. Thanks for the quick approval, @ibnesayeed!

PR #697, opened 2020-07-10
Change mechanism for generating a temporary cdxj file.

### LINKED ISSUE #696: Temporary CDXJ files are not found by replay
Using latest master 63b2805, Python 3.7.7 macOS 10.15.5

```py
ipwb index samples/warcs/5mementos.warc | ipwb replay
Processing WARC records in 5mementos.warc complete
IPWB replay started on http://localhost:5000
 * Serving Flask app "ipwb.replay" (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
```

(access http://localhost:5000/ in a browser)

```
Retrieving URI-Ms from /var/folders/k_/0fg5752j3vl3b814h2qyml_00000gn/T/tmpaai_dd3d.cdxj
[Errno 2] No such file or directory: '/var/folders/k_/0fg5752j3vl3b814h2qyml_00000gn/T/tmpaai_dd3d.cdxj'
(<class 'FileNotFoundError'>, FileNotFoundError(2, 'No such file or directory'), <traceback object at 0x10a098c80>)
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.7/site-packages/ipwb/replay.py", line 579, in showLandingPage
    mementoInfo = calculateMementoInfoInIndex(iFile)
  File "/usr/local/lib/python3.7/site-packages/ipwb/replay.py", line 931, in calculateMementoInfoInIndex
    indexFileContents = get_web_archive_index(cdxjFilePath)
  File "/usr/local/lib/python3.7/site-packages/ipwb/backends.py", line 92, in get_web_archive_index
    response = fetch_local_index(path)
  File "/usr/local/lib/python3.7/site-packages/ipwb/backends.py", line 64, in fetch_local_index
    with open(path, 'r') as f:
```

This _might_ be due to @ibnesayeed adaptation of the temp file to use the native library instead of our own temporary file generation approach. Perhaps the file is not persistent.

machawk1:
Closes #696

### DIFF
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 import random  # For generating a temp file for stdin
 import string  # For generating a temp file for stdin
 import sys
@@ -45,7 +46,8 @@ def checkArgs_replay(args):
         # Write data to temp file (sub-optimal)
 
         fh, args.index = tempfile.mkstemp(suffix='.cdxj')
-        with open(fh, 'w') as f:
+        os.close(fh)
+        with open(args.index, 'w') as f:
             f.write(cdxjIn)
 
         suppliedIndexParameter = True
--- a/ipwb/__main__.py
+++ b/ipwb/__main__.py
@@ -44,10 +44,10 @@ def checkArgs_replay(args):
         random.seed()
         # Write data to temp file (sub-optimal)
 
-        tf = tempfile.NamedTemporaryFile(mode='w', suffix='.cdxj')
-        tf.write(cdxjIn)
-        args.index = tf.name
-        tf.close()
+        fh, args.index = tempfile.mkstemp(suffix='.cdxj')
+        with open(fh, 'w') as f:
+            f.write(cdxjIn)
+
         suppliedIndexParameter = True
 
     proxy = None
--- a/tests/testUtil.py
+++ b/tests/testUtil.py
@@ -57,24 +57,23 @@ def countCDXJEntries(cdxjData):
     return urimCount
 
 
-def startReplay(warcFilename):
+def startReplay(warc_filename):
     global p
-    pathOfWARC = os.path.join(
+    path_of_warc = os.path.join(
         Path(os.path.dirname(__file__)).parent,
-        'samples', 'warcs', warcFilename)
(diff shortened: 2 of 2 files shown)


ipwb/__main__.py
```diff
@@ -44,10 +44,10 @@ def checkArgs_replay(args):
         random.seed()
         # Write data to temp file (sub-optimal)
 
-        tf = tempfile.NamedTemporaryFile(mode='w', suffix='.cdxj')
-        tf.write(cdxjIn)
-        args.index = tf.name
-        tf.close()
+        fh, args.index = tempfile.mkstemp(suffix='.cdxj')
+        with open(fh, 'w') as f:
+            f.write(cdxjIn)
```
ibnesayeed, line 49: They suggests using `os.close(fh)` to properly close files created like this. I would suggest you immediately close the `fh` after creating it, then using your `with open` block on the `args.index`. Also, a similar code is present in the testUtil file, that needs to be fixed as well.

machawk1: > They

who?

Doesn't only opening it in the `with` block close it when the scope is escaped?

> using your with open block on the args.index

What does this mean?

ibnesayeed: > who?

All those smart people on the web.

> Doesn't only opening it in the `with` block close it when the scope is escaped?

Yes, it does, but I am afraid we are attempting to open an existing file handle, not a file name. Opening a file handle works, but I am cautious here because many people suggested that file handle be closed using `os.close(fh)` and not the usual `fh.close()`, I did not investigate why do they say that though.

> What does this mean?

I meant something like this:

```
        fh, args.index = tempfile.mkstemp(suffix='.cdxj')
        os.close(fh)
        with open(args.index, 'w') as f:
            f.write(cdxjIn)
```

Don't forget to import `os` package, if not imported already.

Issue #600, opened 2018-11-28
Revise Replay UI to use CSS grid


machawk1:
In #495 and #502, some issues of the replay user interface were highlighted. In each, the recommendation was to move the UI to use CSS Grid, which was the consensus. No GitHub issue previously existed to explicitly accomplish this. This issue is _that_ issue. :-)

The elements to be included in the main replay grid layout, per the above tickets, will be reduced as some elements are moved to the `/ipwbadmin` endpoint. An expectation is that users that are not running the replay system themselves (i.e., accessing others' replay systems) should not necessarily be able to upload WARCs, know where on the replay's file system indexes are stored, etc.

To be included:
* ipwb logo
* input text field for URI-R entry
* `Search URL in the archive` submit button
* NEW: admin link

This would remove from the main replay interface and move to the admin interface:
* X mementos available
* Everything under Options and Details (currently inclusive of Archive Index(es), Service Worker Version, and Upload WARC titles and corresponding values/buttons.
* IPFS WebUI and IPWB Help links

I do self-question whether the ServiceWorker buttons should be removed, since their intention is to be leveraged by the user viewing the archive (perhaps without admin permissions) and not necessarily the archive maintainer. However, the buttons relating to the SW might be a little too low-level for the casual user, e.g., does someone viewing the archive _really_ need to know the SW details? A "Fix" button or something of the like might be a better solution in lieu to be present on the main replay interface.

@ibnesayeed Please provide feedback on the above, since you have repeatedly suggested we move to use Grid.

ibnesayeed:
We should also include cards (or a brief list) of featured resources on the landing page.

We do not need any buttons or UI to play with the SW, it perhaps was a good idea to have it there for debugging initially while we were still experimenting with it, but this should not be something users should even know about. We just need to do the right thing behind the scene.

machawk1:
The card approach toward displaying the holdings is interesting as an alternative to a list of URIs/titles.

I think there still ought to be something on the replay interface, but much more subtle, to fix issues like cached SWs. Browsers are way too aggressive with caching of SW scripts as the state of the art to allow this behavior to be the culprit of unintended behavior. For instance, I have cleared SWs using devtools in testing yet the cleared one is still used on replay. Switching to a clean Incognito for testing helps but can't be expected for users.

machawk1:
The list of URI-Rs and corresponding datetimes was ported to the `ipwbadmin` page in #712, so it might be time to remove that from the main landing page.

[...]

PR #712, opened 2020-08-25
Provide list of URI-Rs and datetimes. Slight refactoring

### LINKED ISSUE #543: ipwbAdmin: Provide list of URI-Rs and datetimes available

machawk1:
Closes #543

### DIFF
--- a/ipwb/assets/admin.css
+++ b/ipwb/assets/admin.css
@@ -13,4 +13,3 @@ details label.twoRowLabel {height: 2.0em;}
 form input[type=submit] {display: block; clear: both; margin: 0.5em auto 0 auto;}
 h1 {font-size: 24px; margin-bottom: 0.5em; font-family: sans-serif;}
 h1 img {height: 2.0em;}
-ul#uriList li {display: inline;}
--- a/ipwb/assets/admin.css
+++ b/ipwb/assets/admin.css
@@ -11,3 +11,6 @@ details label#daemonStatusLabel {height: 10px; margin-top: 0; vertical-align: to
 details label.twoRowLabel {height: 2.0em;}
 
 form input[type=submit] {display: block; clear: both; margin: 0.5em auto 0 auto;}
+h1 {font-size: 24px; margin-bottom: 0.5em; font-family: sans-serif;}
+h1 img {height: 2.0em;}
+ul#uriList li {display: inline;}
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -578,7 +578,7 @@ def show_admin():
 
     # TODO: determine why summary.htmlCount work propagate to template
     return render_template('admin.html', status=status, indexes=indexes,
-                           uris=uris, htmlCount=htmlCount, summary=summary)
+                           summary=summary)
 
 
 @app.route('/', strict_slashes=False)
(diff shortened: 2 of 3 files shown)


ipwb/assets/admin.css
```diff
@@ -11,3 +11,6 @@ details label#daemonStatusLabel {height: 10px; margin-top: 0; vertical-align: to
 details label.twoRowLabel {height: 2.0em;}
 
 form input[type=submit] {display: block; clear: both; margin: 0.5em auto 0 auto;}
+h1 {font-size: 24px; margin-bottom: 0.5em; font-family: sans-serif;}
+h1 img {height: 2.0em;}
+ul#uriList li {display: inline;}
```
ibnesayeed, line 16: Do we really want to keep the list inline?

ipwb/replay.py
```diff
@@ -575,12 +575,14 @@ def showAdmin():
                'htmlCount': htmlCount,
                'earliest': oldest_datetime,
                'latest': newest_datetime}
+
+    # TODO: determine why summary.htmlCount work propagate to template
     return render_template('admin.html', status=status, indexes=indexes,
-                           summary=summary)
+                           uris=uris, htmlCount=htmlCount, summary=summary)
```
ibnesayeed, line 581: Why are we supplying `uris` and `htmlCount` separately when they are already part of the `summary` object?

machawk1: @ibnesayeed See the comment just preceding 578. There is an issue of the values of summary, namely htmlCount, not being available in the template. This is a separate issue. For now, let's at least get the values displayed then refactor the implementation. :-)

ipwb/assets/admin.css
machawk1: @ibnesayeed I anticipate the design to be heavily refactored, e.g., #600. Inline here is to override the value pulled in from webui.css, which dictates that the values are initially hidden. We can revise this but (Step 1) Get the values display per the GitHub issue.

ipwb/replay.py
machawk1: Resolved in 34882fe42bcd20d8a7b1292d9395fe6524559abd.

ipwb/assets/admin.css
machawk1: Resolved in 34882fe42bcd20d8a7b1292d9395fe6524559abd

ibnesayeed: To undo the effect of an inherited CSS value in this case you could have used `display: list-item;` or `display: initial;`.

ipwb/replay.py
```diff
@@ -575,12 +575,14 @@ def showAdmin():
                'htmlCount': htmlCount,
                'earliest': oldest_datetime,
                'latest': newest_datetime}
+
+    # TODO: determine why summary.htmlCount work propagate to template
```
ibnesayeed, line 579: Are you suggesting that other attributes of the `summary` object are available except a few?

machawk1: The effect is no longer occuring after removal, perhaps a caching issue.

machawk1:
@ibnesayeed Approved on another medium. Merging!

PR #713, opened 2020-08-26
Rename vars to match a consistent, standard style.

machawk1:


machawk1:
Changed to something less cryptic in a975ca3, @ibnesayeed, though the semantics for this variable could be better.

PR #718, opened 2020-09-28
Bugfix daemon

### LINKED ISSUE #716: Update to test with go-ipfs 0.7.0
Released September 22, 2020: https://github.com/ipfs/go-ipfs/releases/tag/v0.7.0

### LINKED ISSUE #717: Daemon CLI argument not implemented
The -d (daemon address) CLI argument isn't implemented, also stated in the help.

I use an external IPFS daemon with an exposed API behind nginx as a reverse proxy. 
Right now it is not possible to use an external Daemon.

gador:
This bugfix adds an implementation for the daemon address and therefore closes #717 
 
it utilizes a settings class (called App) to store settings in.This is necessary, because the flask web app needs to know the daemon address (and can't easily be passed). Also, passing the daemon API address as an argument to every function is cumbersome. 
 
Right now, the settings class only uses the daemon API address, but could be extended to possibly separate the .ipfs/config from ipwb.

Also adds checks for the WebUI to only attempt to  control locally running daemons.


machawk1:
Hi @gador, thanks for the PR! In discussing your additions with @ibnesayeed, we wondered if there was a reason that you used a class to store the attributes in settings.py rather than use it as a simple dictionary? Having to retrieve values through settings.App.get('attr') rather than something like settings['attr'] seems verbose.

Regardless, your changes are very welcomed in removing the need to pass around the daemon multiaddress and toward our move to not have to store the ipwb values in the IPFS config solely for retrieval around the code.

gador:
hi @machawk1 and @ibnesayeed ,
thanks for having a look!

I was looking around how to properly implement settings in python and stumbled upon [this StackOverflow answer](https://stackoverflow.com/questions/6198372/most-pythonic-way-to-provide-global-configuration-variables-in-config-py).

One of the main reasons to use a class for me, is that it avoids global variables. It uses a separate get and set method and is in itself a bit cleaner.

I read the new issue #719 and found the python implementation of multiaddr. I added a way to use the class to validate the entered multiaddr, so we won't have to do it manually. I also added some rudimentary tests.

I'm also working on a way to remove the ipfs config dependency and habe a working version on my pc (and on my github page under [this branch](https://github.com/gador/ipwb/tree/feature_ipfs_config)). 

If this PR with the daemon_address looks alright to you, I'll rebase the feature_ipfs_config branch on it and submit a new PR.

machawk1 (approved):
LGTM. Waiting on @ibnesayeed to review before merging.

machawk1:
Ping @ibnesayeed. This PR was submitted a month ago and is awaiting your approval and/or comment.

machawk1:
Thanks again for your contribution, @gador.

PR #733, opened 2021-06-30
Update testing to use go-ipfs 0.9.0

### LINKED ISSUE #732: Add support in testing for go-ipfs 0.9.0
Released June 22, 2021.

https://github.com/ipfs/go-ipfs/releases/tag/v0.9.0

This has been problematic in the past due to ipfs-http-client (Py library), which had hard-coded version checking, but I believe they changed this function in the last release of the library.

machawk1:
Closes #732


.github/workflows/test.yml
```diff
@@ -24,6 +24,7 @@ jobs:
           - 0.6
```
ibnesayeed, line 24: Should we remove version `0.4` and `0.5` (and perhaps also `0.6`) from our ever-growing test matrix?

machawk1: @ibnesayeed Yes. Would you be ok with keeping 0.7 and above?

ibnesayeed (approved):
LGTM!

Yes, let's stick to the last 3 versions for testing matrix in the future.

PR #742, opened 2021-08-23
Fixes issue of CDXJ TimeMaps returning embedded SURTed URI-Rs in URI-Ms

machawk1:
Closes #741

Issue #180, opened 2017-06-07
Support Link and JSON-formatted TimeMaps


machawk1:
In #105 we describe basic Memento support with a deliverable being a TimeMap payload. For reasons of extensibility, we opted for a CDXJ TimeMap. This is due to the extensibility of the JSON field for each memento/entry, necessary for the IPFS hashes used to retrieve the content by IPWB.

The JSON format should be able to accommodate this format as well but neither JSON nor the more standard Link-formatted TimeMaps have a standard way to represent this necessary information. Supporting Link formatted TimeMaps will make ipwb more interoperable with other Memento tools.

[MemGator](https://github.com/oduwsdl/memgator) writes TimeMaps of all three of these formats, so its output can be used as a model.

machawk1:
CDXJ TimeMap support added in e4a9ee9e5a27feb0158c3ce735e5a0555835b8c9. Tests should still be written for this, as it will likely be a big ding in coverage given the new function.

PR #743, opened 2021-08-23
Add tests for TimeMap retrieval via replay endpoint

machawk1:
Per @ibnesayeed at https://github.com/oduwsdl/ipwb/pull/742#issuecomment-904059560

While a better approach might be to have unit tests for the functions, I believe this addition is a step in the right direction.

Relates to #180, too.


ibnesayeed (approved):
LGTM!

PR #745, opened 2021-10-05
Update to go-ipfs 0.10.0 for testing

### LINKED ISSUE #744: Update tests to use go-ipfs version 0.10.0 (latest)
https://github.com/ipfs/go-ipfs/releases/tag/v0.10.0

Released October 1, 2021

machawk1:
Close #744

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,9 +19,9 @@ jobs:
           - 3.8
           - 3.9
         ipfs:
-          - 0.7
           - 0.8
           - 0.9
+          - 0.10
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os }} Py-${{ matrix.python }} IPFS-${{ matrix.ipfs }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -19,9 +19,9 @@ jobs:
           - 3.8
           - 3.9
         ipfs:
-          - 0.7
           - 0.8
           - 0.9
+          - 0.10
```
ibnesayeed, line 24: YAML parser reads it as a decimal number and removes any trailing zeros:

```suggestion
          - "0.10"
```

ibnesayeed (changes requested):

.github/workflows/test.yml
machawk1: Good catch, @ibnesayeed! I was just reading about YAML's interpretation of Python 3.10, which would have a similar issue.

ibnesayeed (approved):
LGTM!

PR #749, opened 2021-12-15
Update tests to also use Python 3.10 re:#748

### LINKED ISSUE #748: Update automated tests to use Python 3.10
The testing GitHub Action currently uses versions 3.7, 3.8, and 3.9. Let's first add 3.10, if available, and if it ever becomes problematic, eventually deprecate 3.7.

machawk1:
Closes #748


ibnesayeed (approved):
LGTM!

How many past Python versions do we want to continue to test against?

machawk1:
@ibnesayeed That might need to be contingent in the prevalence of the older versions as reported on surveys. I have no strong feelings here. If you do, let them be known.

ibnesayeed:
I am okay with keeping them around. It is just that the test matrix grows bigger and takes longer to test. No big deal though!

PR #750, opened 2022-02-11
Fix the broken link to the PDF in the README

machawk1:
A better solution would be to push the PDF to this repo but then we would have to deal with rights issues. This solves the problem for now of ODUCS pages becoming inaccessible.


ibnesayeed (approved):
LGTM!

@machawk1 could you look at the failed tests please?

PR #757, opened 2022-03-30
Update go-ipfs in tests to version 0.12.1

### LINKED ISSUE #756: Update tests to use latest go-ipfs 0.12.1
Release March 19, 2022, https://github.com/ipfs/go-ipfs/releases/tag/v0.12.1

machawk1:
Closes #756

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,7 +22,7 @@ jobs:
         ipfs:
           - "0.10"
           - "0.11"
-          - "0.12"
+          - "0.12.1"
     runs-on: ${{ matrix.os }}
     name: ${{ matrix.os }} Py-${{ matrix.python }} IPFS-${{ matrix.ipfs }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         ipfs:
           - "0.10"
           - "0.11"
-          - "0.12"
+          - "0.12.1"
```
ibnesayeed, line 25: Is this change required? I have added semver resolution in the corresponding workflow, so it should resolve to the latest point release without version pinning to the third point.

machawk1: Let me revert this to just use 0.12 here then inspect the Action.

machawk1: @ibnesayeed You were correct -- the 0.12 specification here was correctly resolved to 0.12.1.

ibnesayeed (approved):
LGTM!

PR #758, opened 2022-04-04
Move to use port 2016 as the default instead of port 5000

### LINKED ISSUE #752: Change default replay port from 5000
A recent macOS update uses port 5000 by default for AirPlay Receiving. This is built into the system and turned on by default. While another port could easily be expressed at runtime, let's pick a new default.

machawk1:
Closes #752


ibnesayeed (approved):
LGTM!

PR #766, opened 2022-05-08
Add host flag to replay

### LINKED ISSUE #763: Allow declaration of a custom hostname on launching the replay
Per #761, independent of the proxy flag, suggested by @ibnesayeed

machawk1:
Closes #763

### DIFF
--- a/ipwb/replay.py
+++ b/ipwb/replay.py
@@ -1047,15 +1047,15 @@ def get_cdxj_line_binary_search(
     return line_found
 
 
-def start(cdxj_file_path, proxy=None, port=IPWBREPLAY_PORT):
+def start(cdxj_file_path, proxy=None, host=IPWBREPLAY_HOST, port=IPWBREPLAY_PORT):
     host_port = ipwb_utils.get_ipwb_replay_config()
     app.proxy = proxy
 
     # Retain port for subsequent runs
-    ipwb_utils.set_ipwb_replay_config(IPWBREPLAY_HOST, port)
+    ipwb_utils.set_ipwb_replay_config(host, port)
 
     if not host_port:
-        host_port = (IPWBREPLAY_HOST, port)
+        host_port = (host, port)
 
     # This will throw an exception if daemon is not available.
     ipwb_utils.check_daemon_is_alive()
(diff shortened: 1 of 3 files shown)


ipwb/replay.py
```diff
@@ -1067,7 +1067,7 @@ def start(cdxj_file_path, proxy=None, port=IPWBREPLAY_PORT):
         print((f'IPWB replay started on '
                f'http://{host_port[0]}:{host_port[1]}'))
 
-        app.run(host='0.0.0.0', port=host_port[1])
+        app.run(host=host_port[0], port=host_port[1])
```
ibnesayeed, line 1070: I think it is better to leave it `0.0.0.0`, if not then we should add a special case where we define things in utils.

ipwb/util.py
```diff
@@ -33,7 +33,7 @@
 # or '/ip4/{ipaddress}/tcp/{port}/http'
 # or '/ip6/{ipaddress}/tcp/{port}/http
 
-IPWBREPLAY_ADDRESS = 'localhost:2016'
+IPWBREPLAY_ADDRESS = '0.0.0.0:2016'
```
ibnesayeed, line 36: The `0.0.0.0` is a "listen to all interfaces" address, it can't be resolved from the client side.

ibnesayeed (changes requested):

ipwb/replay.py
ibnesayeed: It really depends on the purpose of the `host` parameter. If it is to be used for formatting FQDN URIs then it really does not matter which interface the service is listening to, as long as it is reachable, which means running it on `0.0.0.0` will do, irrespective of what hostname is used.

ipwb/util.py
ibnesayeed: If you want host parameter to be customizable, then add a special case to see if it is set to `0.0.0.0` then for all URI rewriting and FQDN URI purposes use `localhost` (or the value in the `proxy` parameter).

ipwb/replay.py
machawk1: Ok, reverted in [81ae262](https://github.com/oduwsdl/ipwb/pull/766/commits/81ae2622a261bca6b3843e48bb8ff99e85250a41), but the specified host still does not appear in the TimeMaps. I will move this PR to WIP for now until that issue is resolved for this branch.

Issue #710, opened 2020-08-13
Support indexing WACZ files


machawk1:
Web Archive Collection Zipped (WACZ) Format, https://github.com/webrecorder/wacz-format (MIT, potentially reusable)

Example of MDN WACZ at https://twitter.com/webrecorder_io/status/1293730279824089088

https://dh-preserve.sfo2.cdn.digitaloceanspaces.com/webarchives/mdn.wacz (1.6GB)

machawk1:
Preliminary support added in [779978a](https://github.com/oduwsdl/ipwb/commit/779978ae6480dd5f677c9e5a5a97dbb7c0b3546d). WACZ detection should be improved but importing py-wacz incurs others dependencies due to pywb coupling.

machawk1:
Also, `is_zipfile()` fails on WACZ files due to the magic number (signature) not matching that of a ZIP file.

ikreymer:
> ```
> wacz create -o ./samples/wacz/my-collection.wacz ./samples/warcs/5mementos.warc ./samples/warcs/froggie.warc.gz ./samples/warcs/salam-home.warc
> ```

The command should include a `-f` before the WARCs files now (and should have better arg validation probably)

machawk1:
@ikreymer Thanks for your proactive feedback here. I ran:

`wacz create -o ./samples/wacz/my-collection.wacz -f ./samples/warcs/5mementos.warc ./samples/warcs/froggie.warc.gz ./samples/warcs/salam-home.warc` and a 79 KB file
[my-collection.wacz.zip](https://github.com/oduwsdl/ipwb/files/8710885/my-collection.wacz.zip) (.zip added only for GitHub upload) was generated. This WACZ does not cause any URLs to be recognized in replayweb.page.

wacz 0.4.6 installed via pypi, macOS 12.3.1, Python 3.10.4

machawk1:
Per a discussion w/ Mark G. @ IA, WACZ is supported at web-beta.archive.org/save for those with a "beta" account (which I have).

PR #770, opened 2022-05-24
Add WACZ support

machawk1:


machawk1:
Pending a test passing and probably some better documentation, but I would appreciate your feedback on the diff here, @ibnesayeed.

ipwb/util.py
```diff
@@ -336,3 +340,28 @@ def check_for_update(_):
...
+
+def get_warc_paths_in_wacz(wacz_path):
+    with ZipFile(wacz_path) as z:
+        return [w for w in z.namelist() if w.startswith('archive/')]
+
+
+def extract_warcs_to_disk(wacz_path, warc_paths):
+    for warc in warc_paths:
+        with ZipFile(wacz_path) as z:
+            z.extract(warc)
```
ibnesayeed, line 360: I would expect a post-indexing cleanup process where these extracted WARCs are removed after they are indexed.

ipwb/util.py
```diff
@@ -336,3 +340,28 @@ def check_for_update(_):
...
+
+
+def extract_warcs_from_wacz(wacz_path):
```
ibnesayeed, line 363: Would it not be easier to unpack the whole WACZ file (with force overwrite), ideally in a new temp folder, instead of extracting WARCs one by one? That way, the whole `archive/` folder can be added in the WARCs targets to scan from.

ipwb/indexer.py
```diff
@@ -123,6 +124,18 @@ def index_file_at(warc_paths, encryption_key=None,
...
+    warc_paths_to_remove = []
+    for warc_path in warc_paths:
+        if is_wacz(warc_path):
+            warc_paths_to_append += extract_warcs_from_wacz(warc_path)
+            warc_paths_to_remove.append(warc_path)
+
+    # Manipulate list of WARCs extracted from WACZ
+    for ptr in warc_paths_to_remove:
+        warc_paths.remove(ptr)
+    warc_paths = warc_paths + warc_paths_to_append
```
ibnesayeed, line 137: I have noticed that in some WACZ files sometimes there are `.warc` files, which are actually `.warc.gz` files. I hope `warcio` transparently handles those.

ibnesayeed (commented):
I would prefer we handle it by externally extracting WACZ files and then indexing those WARCs the usual way until we have a built-in WARC record iterator in py-wacz (an idea that I discussed with @ikreymer already), at which point this change will become obsolete.

That said, I do not have any strong objections against this PR. I have added some inline comments for potential improvements though.

ipwb/indexer.py
machawk1: This would make a good test case.

ipwb/util.py
machawk1: A post-indexing cleanup process has been implemented in [2566b7c](https://github.com/oduwsdl/ipwb/pull/770/commits/2566b7cfc47667c78de532ee7d9bcbed340ac16c). Thanks for the suggestion, @ibnesayeed

ipwb/indexer.py
ibnesayeed: I think we do not need to define `warcs_to_cleanup_post_indexing` and use the existing `warc_paths_to_append` in its place.

ibnesayeed: I think we should rename `warc_paths_to_remove` to something like `wacz_paths`, because the use of the term `remove` can now be confusing if it means the path will be removed from being processed or the file will be removed (especially, after introducing the cleanup function that removes files).

ipwb/util.py
ibnesayeed: I think, we should create a temporary folder and extract files into that, then remove the whole temporary folder for cleanup. Some reasons to use a temporary folder:

* The process might not have write permissions in the folder where things are extracted
* If there were WARC files with duplicate names in more than one WACZ files, they might get overwritten, and if the name conflicts with a WARC file outside of the WACZ file, that file will be deleted during cleanup, which is unintended

machawk1: Good catch. Do you suggest we rename the files to some temporary filename and put it in a temporary location? This might help to prevent conflicts.

Extract (from wacz) --> (destination):
/archive/mywarc1.warc --> /tmp/js9gar/jkldfsdf.warc
/archive/mywarc2.warc --> /tmp/js9gar/234vdsf.warc
/archive/mywarc3.warc --> /tmp/js9gar/0sdf89a.warc

Then delete files at (destination) or more simply the /tmp/js9gar directory.

ipwb/indexer.py
machawk1: Renaming and conversion to temp directories introduced in [e4b9998](https://github.com/oduwsdl/ipwb/pull/770/commits/e4b9998f82f7cb75f63ed808ec5080c0e7449074).

ipwb/util.py
machawk1: Complete in [e4b9998](https://github.com/oduwsdl/ipwb/pull/770/commits/e4b9998f82f7cb75f63ed808ec5080c0e7449074)

ipwb/util.py
machawk1: Resolved with [e4b9998](https://github.com/oduwsdl/ipwb/pull/770/commits/e4b9998f82f7cb75f63ed808ec5080c0e7449074)

ipwb/indexer.py
```diff
@@ -171,6 +186,8 @@ def index_file_at(warc_paths, encryption_key=None,
     cdxj_metadata_lines = generate_cdxj_metadata(cdxj_lines)
     cdxj_lines = cdxj_metadata_lines + cdxj_lines
 
+    cleanup_warc_files_extracted_from_wacz(warc_paths_to_append)
```
ibnesayeed, line 189: Not a big issue, but I think the temporary folders created by the `mkdtemp()` call will continue to exists (until cleaned up by the OS) because only the files inside them are deleted, not the folders themselves.

machawk1: The docs say that the creator is responsible for the deletion, so I think we should handle this. Given each WARC gets a new temp directory, it might be better to just retain the copy of this directory path and delete it along with its contents instead of deleting the WARC then the directory, which would require tracking the directory path, too. Which approach would you rather be implemented, @ibnesayeed?

ibnesayeed: > which would require tracking the directory path, too

Not really! It is possible to get the path of the directory if the path of a file is known that it contains.

That said, I would perhaps preferred not holding onto the list of WARC files, instead, operate on each WARC as we discover them, whether those are regular WARC files or those extracted from WACz files. I would deal with one file at a time and loop over for the next one.

machawk1: @ibnesayeed This seems like it requires a revamp outside of the scope of this GH issue/PR. I agree that dealing with one WARC at a time would likely be more computationally optimal.

ibnesayeed: I understand that it would require change in the workflow. When done, it would be more space efficient as not all the WARCs need to be extracted from WACZ files upfront, duplicating them on the disk, before processing them.

It is okay to leave it as things are right now and get back to this when we have a WARC record iterator for the WACZ files, when most of these changes will be rendered useless.

ikreymer:
While its nice to see more support for WACZ, I am wondering about how you're seeing support the WACZ format in this implementation.

It seems like this PR is simply treating WACZ as just ZIP containing some WARC files. 
If that is the case, why not just add documentation for users to import a WACZ file by using:
```
unzip somefile.wacz
ipwb index archive/*.warc.gz
```

But a WACZ file is more than just a zip of WARC files! More importantly, it has a builtin CDXJ in a specific format, which allows for fast access and does not require reindexing of the WARCs. It also has metadata, fixity information (payload and full record digests), and optional signature. There is a specific, extending the frictionless data package, and validated by py-wacz.

But if everything but the WARCs are being discarded, I wonder if that should be considered 'supporting WACZ' or just 'importing WARCs from WACZ file'.

Can you see a way for ipwb to support more of the WACZ format, such as reusing the existing CDXJ indexes perhaps?

In some ways, the current approach is a lossy transformation: WACZ files can already be hosted and used directly from IPFS. By importing into ipwb, you're taking the WARCs from the WACZ, reindexing them, and putting just the WARCs onto IPFS in a slightly different way, while creating a new index that is not stored on IPFS, and discarding other data (metadata, signatures, etc...). I wonder if ipwb could do something else to better leverage the properties of the WACZ format?

machawk1:
@ikreymer This initial implementation is base support for WACZ. We are hoping to do more with the format in the future beyond treating it as a container for WARCs.

Reusing the CDXJ in the WACZ for ipwb's own generated CDXJ index could be interesting. We have not thought about it much yet.

@ibnesayeed likely has some ideas / 2¢. Thanks for your input!

ibnesayeed:
> It seems like this PR is simply treating WACZ as just ZIP containing some WARC files.
> If that is the case, why not just add documentation for users to import a WACZ file by using:

@ikreymer that would have been my preference as well (as noted in a comment earlier https://github.com/oduwsdl/ipwb/pull/770#pullrequestreview-991912563), but @machawk1 wanted a more built-in approach, hence this PR.

> But if everything but the WARCs are being discarded, I wonder if that should be considered 'supporting WACZ' or just 'importing WARCs from WACZ file'.

In that sense IPWB does not even support WARC files, let alone WACZ files. During the indexing and ingestion process, it iterates over each WARC record, splits the headers and payload to store them as small IPFS objects to be used at the time of replay. This is why we are interested in an iterator over the WARC records and not the container (of the container of the container) format.

> Can you see a way for ipwb to support more of the WACZ format, such as reusing the existing CDXJ indexes perhaps?

We can leverage the CDXJ file packaged inside the WACZ file, but not for replay, only as a means to iterate over WARC records during indexing/ingestion. However, for large WACZ files this approach might be slow because the reader pointer will be jumping around to see to different locations if the CDXJ file is not sorted in the order those records are preserved. Moreover, if the built-in CDXJ excludes certain WARC record types (which might not be the case, but worth mentioning here) then those will not be reachable.

> In some ways, the current approach is a lossy transformation: WACZ files can already be hosted and used directly from IPFS. By importing into ipwb, you're taking the WARCs from the WACZ, reindexing them, and putting just the WARCs onto IPFS in a slightly different way, while creating a new index that is not stored on IPFS, and discarding other data (metadata, signatures, etc...). I wonder if ipwb could do something else to better leverage the properties of the WACZ format?

I am not sure if I would call it lossy, because the goal of IPWB is different from some of the other web archiving systems. It operates on the atomic level of records while gluing related pieces together for replay. IPWB was not modeled to interact with WARC/WACZ or any other container/bundle format at the time of replay. For long term preservation, one is encouraged to retain WARC/WACZ files with as much provenance and metadata as possible. IPWB can be improved to incorporate more metadata followed by re-ingesting WARC files from the cold storage.

machawk1:
@ibnesayeed I would like your re-review here, as I added some logic to retain the temp paths and remove them as required so as to not have side effects. Inferring these _might_ be unreliable, as the WARCs extracted from a WACZ are stored at {unique_temp_path}/archive/warcname.warc.gz rather than simply {unique_temp_path}/warcname.warc.gz.

If you would rather we change this to the latter scenario, which should not cause a clash issue due to the unique paths, let me know and I can cleanup the logic.

Otherwise, the changes here meet the requirements of #710. The latter scenario can always come in with refactoring.

ipwb/indexer.py
```diff
@@ -123,6 +126,21 @@ def index_file_at(warc_paths, encryption_key=None,
     for warc_path in warc_paths:
         verify_file_exists(warc_path)
 
+    # Extract WARCs from any WACZ files
+    warc_paths_to_append = []
+    wacz_paths = []
+    for warc_path in warc_paths:
+        if is_wacz(warc_path):
+            (w_paths, dirs_to_cleanup) = extract_warcs_from_wacz(warc_path)
```
ibnesayeed, line 134: The `dirs_to_cleanup` here is overwritten in each loop, so at the end it will only hold the reference to the temporary dirs of the last WACZ file (unless I am missing something) for cleanup.

ibnesayeed (commented):
I feel like we have some unnecessary code because we are doing things the complicated way. For example, we are maintaining list of extracted WARC files from WACZs to be removed later, then we also maintain list of temporary directories to be removed, though removing the later will automatically remove its contents.

I am not sure if merging this PR would be a good idea as it brings a lot of complexities that will be difficult to maintain and reason about and at some point they might become useless (if and when we have WARC record iterator built in the WACZ library).

ipwb/indexer.py
machawk1: Good catch on `dirs_to_cleanup` not being retained.

machawk1: I think the logic of extracted WARCs and temp directories can be simplified. There is a bit too much disk maintenance. The general gist is that paths to WACZ files could be inter-mingled with WARCs here and thus the WARCs extracted from the WACZ files need to be removed but _not_ the WARCs that were passed in.

machawk1:
- [ ] Change process processing to handle on WARC at a time from extraction, ingestion, extraction, `ipwb magic`, deletion before moving on to next WARC in the WACZ.

